PR #4, opened 2023-12-12
Replace unsafe id generator with nanoid

174n:
This PR replaces Math.random-based id generator with [nanoid](https://github.com/ai/nanoid) which looks like this:

```
ZA4xhzA2RB5pU_2mdZDp9
```

It has multiple security benefits:

  * Longer ids which make brute force attacks basically impossible
  * Safe randomness
  * Collision resistance

I also added .env.example to save some time for future contributors


scastiel:
Thanks for the contribution! Indeed it makes much more sense to use `nanoid`

PR #11, opened 2023-12-15
Add splitmode and shares to expenses

### LINKED ISSUE #6: Split unevenly
It would be nice to be able to split an expense unevenly:
- by percentage (e.g. Alice pays 50%, Bob 25% and Charlie 25%)
- by shares (e.g. Alice pays 2 share, Bob one and Charlie one)
- by amount (e.g. Alice pays 20$,  Bob pays 10$ and Charlie 5$)

scastiel:
Fixes #6

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -34,6 +34,7 @@
         "react-hook-form": "^7.47.0",
         "tailwind-merge": "^1.14.0",
         "tailwindcss-animate": "^1.0.7",
+        "ts-pattern": "^5.0.6",
         "uuid": "^9.0.1",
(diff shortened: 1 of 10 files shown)


ChristopherJohnston:
I get an undefined error on the match method in expense-form.tsx. Was able to rectify by changing:

line 196: defaultValue={field.value} -> defaultValue="EVENLY"

line 356: match(form.getValues().splitMode) -> {match(form.getValues().splitMode ?? "EVENLY")

I see that the "default" value is set in schema but this doesn't appear to be honoured.

PR #16, opened 2024-01-03
Add activeUser for default payer per group

ankitbahl:
This change adds an active user option to group settings, which can be set per group. This is only saved in the devices localStorage, and therefore can be set per device.

### DIFF
--- a/src/app/groups/[groupId]/expenses/expense-list.tsx
+++ b/src/app/groups/[groupId]/expenses/expense-list.tsx
@@ -6,7 +6,7 @@ import { Participant } from '@prisma/client'
 import { ChevronRight } from 'lucide-react'
 import Link from 'next/link'
 import { useRouter } from 'next/navigation'
-import { Fragment } from 'react'
+import { Fragment, useEffect } from 'react'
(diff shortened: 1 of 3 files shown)


scastiel:
Hi @ankitbahl, thanks a lot for the contribution, it looks great!

A couple of remarks about the PR:
- Could you update it with the origin repo and resolve the conflicts (it’s nothing too bad 😉)
- Could you make the local storage contain a participant ID instead of their name?
- I think there is a tiny issue when I add a new participant to a group and want to select it as the active user (the option is _New_ instead of the participant’s name)

Just tell me if you don’t have time (or don’t want) to make the changes, I will gladly take it from here if you prefer 😉

ankitbahl:
Appreciate you taking a look ☺️ Great suggestions, I'll make those changes later today 🙂

ankitbahl:
Realized there was one more bug, should be good to look at now

scastiel:
Amazing, thanks @ankitbahl!

ChristopherJohnston:
Awesome addition!

Issue #17, opened 2024-01-03
Store and display date of transaction


LIRIKKER:
Would be great to have the ability to add the date of the transaction onto the Expense.
With this the homepage of the group should then be sorted by date.
This should also be broken down by months but still remain on the same page.
i.e

January 2024
3rd - Transition 1- £50
2nd - Transition 2 - $23
1st - Transition 3 - £45

Break

December 2023
31st - Transition - 4 £44
21st - Transition - 5 £123
4th - Transition - 6 £12

Break

November 2023

scastiel:
Date is now stored and displayed (#24), I created a separate issue (#27) to group expenses by date.

PR #26, opened 2024-01-08
Add Expense Date

ChristopherJohnston:
Add expense date field, defaulting to today's date. This is assignable in the expense form.
Group expenses list will sort by this date in descending order

This resolves issue #17


scastiel:
Amazing, thanks @ChristopherJohnston! 👏 

I took the liberty to change the field description and how the date is displayed. After the PR is deployed, I’ll run a query to update all `expenseDate`s to the expense creation date (which I think makes sense for existing expenses).

PR #28, opened 2024-01-08
Assign categories to expenses

ChristopherJohnston:
add a categories table
add seed script to populate with some categories (mimics the splitwise category list)
add category relation to expense table,
add ability to select category in the create/edit forms


ChristopherJohnston:
Re-submitted this implementation, merged with the expenseDate and uneven-split changes.

The intent here is to follow the split wise categories so it's easier to import from a CSV.

Happy to discuss the implementation.

scastiel:
@ChristopherJohnston looks great!

I had some trouble to make the migration work, so I changed it a bit, to insert categories into the database as part of the migration (instead of using a seed). It makes more sense to me, what do you think?

scastiel:
Closes #35

ChristopherJohnston:
It looks fine as the migration. I'm now wondering if it's easier and better just to have a hard-coded list of categories to save the additional overhead from database queries - particularly if the categories are going to be fixed.

ChristopherJohnston:
> @ChristopherJohnston looks great!
> 
> I had some trouble to make the migration work

for reference I took the seed script concept from https://planetscale.com/blog/how-to-seed-a-database-with-prisma-and-next-js

scastiel:
> I'm now wondering if it's easier and better just to have a hard-coded list of categories to save the additional overhead from database queries - particularly if the categories are going to be fixed.

That is a fair point, maybe it’ll be easier to deal with hardcoded categories.

Max-TheCat:
> > I'm now wondering if it's easier and better just to have a hard-coded list of categories to save the additional overhead from database queries - particularly if the categories are going to be fixed.
> 
> That is a fair point, maybe it’ll be easier to deal with hardcoded categories.

Since the implementation is pretty much done, I'd keep it in the DB as it keeps things extensible. That's how I was about to implement it anyways :laughing:

ChristopherJohnston:
Yep it's not a bad implementation as it is - just need to be careful of the impact on api calls on Vercel as volume increases. I would think it's fine for now

scastiel:
Thanks @Max-TheCat for the input! I’ll merge the PR as is. Worst case scenario we migrate to a hardcoded list, which wouldn’t be too much work 😉

PR #31, opened 2024-01-08
Fix create expense form when clicking reimbursement link in balances page

ChristopherJohnston:
Upon clicking "Mark as Paid" in the group balances tab, this redirects to the create expense page with argument "reimbursement".
The value for splitMode is not set in this instance so the UI renders the "shares" controls

To fix, explicitly set splitMode: 'EVENLY' for the reimbursement scenario.


scastiel:
🙌

PR #37, opened 2024-01-10
add new field "paid at"

arnimhanke:
Hi,
i added a new field to save the date, when the expense has been paid.
I am happy if you have any feedback, so we can work together at this project!


arnimhanke:
Oh, i see that this feature was added two days ago :-D
So i will  close my PR.

PR #41, opened 2024-01-13
POC Encrypt/Decrypt expense title on client side

### LINKED ISSUE #29: Mark a group as favorite
Marking a group as favorite (stored locally) should make it appear at the top of the recent groups list.

### LINKED ISSUE #33: Expense "Paid by X on" dates showing incorrect
Dates appear 1 day previous on the 'dashboard' view. Clicking to open an individual expense will show the correct date:

image

^Incorrect

image

^Correct

### LINKED ISSUE #38: Remove group from recent groups list

ChristopherJohnston:
Proof of concept outlining SHA-256 encrypt/decrypt of expense title on the client using libp2p/keychain and libp2p/cms, storing the private key in IndexedDb using datastore-idb

NB the encrypt/decrypt calls are quite slow so probably not a practical solution!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,8 @@
       "hasInstallScript": true,
       "dependencies": {
         "@hookform/resolvers": "^3.3.2",
+        "@libp2p/cms": "^2.0.2",
+        "@libp2p/keychain": "4.0.5",
         "@prisma/client": "5.6.0",
         "@radix-ui/react-checkbox": "^1.0.4",
         "@radix-ui/react-collapsible": "^1.0.3",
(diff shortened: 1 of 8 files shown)


ankitbahl:
This looks like RSA (Public key crypto) - For basic encryption, a symmetric key encryption methodology should be used. For example AES. This will likely speed up those calls (Symmetric key cryptography is generally much faster)

Doesn't JS already have build in crypto functionality? Not sure how performant it is comparatively.

ChristopherJohnston:
It's using CMS, which utilises an RSA private key for encryption. This particular proof of concept makes use of libp2p/cryptography which stores the keys in an IndexedDB within the browser's local storage.

The slowness I'm now attributing to the time it's taking for the UI graph to rebuild rather than the individual encrypt/decrypt calls.

That said, I have no real preference to which encryption method is used - this really was an exercise to see if it could be done without significant changes.

ChristopherJohnston:
closing this for now as it's more of a proof of concept - will revisit at a later date

PR #48, opened 2024-01-16
Group expenses my date

acuteengle:
Screen Shot 2024-01-16 at 23 58 43

### DIFF
--- a/src/app/groups/[groupId]/expenses/expense-list.tsx
+++ b/src/app/groups/[groupId]/expenses/expense-list.tsx
@@ -16,6 +16,21 @@ type Props = {
   groupId: string
 }
 
+function getGroupedExpensesByDate(expenses) {
+  return expenses.reduce((result, expense) => {
+    const dateString = formatDate(expense.expenseDate)
+    result[dateString] = result[dateString] ?? []
+    result[dateString].push(expense)
+    return result
+  }, {})
+}
+
+function getOrderedDates(dates) {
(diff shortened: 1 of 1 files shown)


acuteengle:
> There are some TypeScript errors in your code, could you please fix them? You can run npx tsc --noEmit to see them if you can’t see them in your IDE.

Ah thanks! Done!

I also ran `npx prettier . --write` since I noticed you have a CI check for that as well

src/app/groups/[groupId]/expenses/expense-list.tsx
```diff
@@ -16,6 +17,51 @@ type Props = {
...
+  LAST_YEAR: 'Last year',
+  OLDER: 'Older',
+}
+
+function getDifferenceInDays(earlierDate: Date, laterDate: Date) {
+  const differenceInTime = laterDate.getTime() - earlierDate.getTime()
+  return Math.round(differenceInTime / (1000 * 3600 * 24)) // convert milliseconds to days
+}
+
+function getExpenseGroup(date: Dayjs, today: Dayjs) {
```
Comment (acuteengle) [inline src/app/groups/[groupId]/expenses/expense-list.tsx:34]: Originally I had this:
Screen Shot 2024-01-17 at 10 10 42

But then I realized that the headers were then misleading so I have to actually check if the expense date is in the same week, or month, or year, etc. Because of that, I had to utilize the `dayjs` package

acuteengle:
@scastiel Ready for another look!

Screen Shot 2024-01-17 at 10 13 49

PR #57, opened 2024-01-25
Production target for Dockerfile

justcallmelarry:
I modified the Dockerfile in order to allow for both `development` (current functionality) and `production` (built next app) versions, this allows self hosters like myself to run the app in production mode, which greatly improves the resources used, especially memory usage (~1gb to ~100mb for my setup).

I also added:
* Modularity to the scripts regarding docker containers to be able use both these targets.
* New package.json target for starting up the dev container rather than the prod container.
* Instructions to README on how to run both versions.

## BREAKING CHANGES
Note that this PR introduces two breaking changes:
* The default container to build is the `production` build, not the `development` build, you will need to use a different command than earlier in order to build the `development` build (previously built images will still run like before, though).
* I moved the default volume for the database, as it was currently placing them in `/var/lib/postgresql/data`, I changed this to use a relative path (`./postgres-data`, same as the script starting the local db when not running in container). Anyone who used the `compose.yml` as is will need to move/copy the data from the previous location to the new one. Or to just update their versions of the compose file.

I made the breaking changes as I though these are saner defaults, but I am willing to change them back (and reverse the `build-image`/`build-image-dev` commands to `build-image`/`build-image-pro`, just let me know if this is the case.

Would recommend squash-merging in case of merge, since commit history got pretty messy 😬


scastiel:
Thanks @justcallmelarry, I tested it and it seems to work perfectly!

Naive question (I’m not used to Docker): is there a point to keeping the `dev` workflow? It seems that even if the application is running in dev mode, the dev features such as hot reload don’t work anyway (as the files are copied at build time). Or am I missing something?

justcallmelarry:
@scastiel hmm, I was thinking of leaving it since it was done this way for some reason to start with

Hot reloading doesn't work unless you volume the code into the container at runtime, then if should behave just like if you were running `npm run dev` locally

but i now notice that I dont get hot reloading when doing that either… so maybe you are correct in that the dev workflow is redundant

ill remove it and just keep the production workflow in its place

scastiel:
I think it’s safe to remove the dev workflow, but curious to get @ChristopherJohnston and/or @Max-TheCat’s point of view about it in case we’re missing something.

Dockerfile
```diff
@@ -1,17 +1,36 @@
...
 RUN apt update && \
     apt install openssl -y && \
     apt clean && \
     apt autoclean && \
-    apt autoremove && \
-    npm install --ignore-scripts && \
-    npm install -g prisma
+    apt autoremove
+
+RUN npm ci --ignore-scripts
```
Max-TheCat, line 13: I'd recommend doing everything in a single RUN directive not to add too many layers to the image.

Max-TheCat:
@scastiel I'm not familiar with next.js build/deployment process so it's possible my image is lacking important stuff. I'm all in for the addition of a prod image.

For the hot reload issue, there might be a port we need to expose in the dev image for it to work. Maybe using Viscose dev environment with the dev image would work also. I'd keep the dev image also for the time being.

@justcallmelarry If you can pack your cmds in a single RUN directive I'm 100% good with your changes. It also works as it is but may produce larger images.

justcallmelarry:
Ok, so! I figured out what I was doing wrong (I'm a Python/Go developer mainly, and just a happy amateur when it comes to JS/TS), and it seems that you need to visit the site first (and have it compiled) before hot reloading takes effect.

When adding the following volume to the app in the compose file hot reloading works as expected:
```yaml
volumes:
    - ./src:/usr/app/src
```

So this is basically the way you would want to add the volumes for everything you wish to hot reload.

As this was not included in the original PR for a containerised version, I also didn't add the volumes now (as they are kinda useless for the prod image).

If @Max-TheCat (who made the original dockerization) is OK with not running the app in development mode, I can remove the development workflow all-together, as it's not very easy to just run it in development mode in the current setup anyway.

Otherwise I can keep the development target and add a comment about the volumes in the development target part of the readme? Let me know and I'll update.

Dockerfile
justcallmelarry: I also tried to get a version going where I just copy over the built files from the `build` image target to the `production` image target, but my lack of knowledge of how and what Next files i need in order to run it prohibited me from ending up with a slimmer image.

I might look at it again when I have more time, though.

Max-TheCat:
> Ok, so! I figured out what I was doing wrong (I'm a Python/Go developer mainly, and just a happy amateur when it comes to JS/TS), and it seems that you need to visit the site first (and have it compiled) before hot reloading takes effect.
> 
> When adding the following volume to the app in the compose file hot reloading works as expected:
> 
> ```yaml
> volumes:
>     - ./src:/usr/app/src
> ```
> 
> So this is basically the way you would want to add the volumes for everything you wish to hot reload.
> 
> As this was not included in the original PR for a containerised version, I also didn't add the volumes now (as they are kinda useless for the prod image).
> 
> ~If @Max-TheCat (who made the original dockerization) is OK with not running the app in development mode, I can remove the development workflow all-together, as it's not very easy to just run it in development mode in the current setup anyway.~
> 
> ~Otherwise I can keep the development target and add a comment about the volumes in the development target part of the readme? Let me know and I'll update.~
> 
> EDIT: Sorry Max, saw now that you'd want to keep the dev image. I'll add some info to the README instead!

I'm also ok if we remove it. I'd let @scastiel make the call though. Maintaining a single image sure is simpler.

scastiel:
> I'm also ok if we remove it. I'd let @scastiel make the call though. Maintaining a single image sure is simpler.

I suggest we remove the dev workflow, at least until someone expresses a need for it. Maybe I’m biased but if you want run the app for development, I can’t see why you’d want to use Docker instead of a plain old `npm install && npm run dev`.

ChristopherJohnston:
sorry for the delay - I did like the idea of the docker dev workflow as it meant not having to run the extra ./scripts/start-local-db to get a PostgresDB image up. Having said that I am quite agnostic about using npm run dev vs docker-compose up.

The current trend for dev workflows in vs code seems to be to use dev containers as it isolates the environment and provides any tooling pre-installed. That might be something to explore.

justcallmelarry:
> > I'm also ok if we remove it. I'd let @scastiel make the call though. Maintaining a single image sure is simpler.
> 
> I suggest we remove the dev workflow, at least until someone expresses a need for it. Maybe I’m biased but if you want run the app for development, I can’t see why you’d want to use Docker instead of a plain old `npm install && npm run dev`.

For this project that might be true 😄 Docker development has many pluses, included but not limited to:
* not having to install certain dependencies if needed (like anything that has to do with image handling) on your local machine.
* having multiple services including infrastructure services (think things like Kafka, RabbitMQ, DynamoDB, etc) run in the same network
* having another version of node installed, avoiding inconsistencies between the versions (i know this can be handled by stuff like npx etc, but anyway)

I do think that the development image for this project i as of now somewhat redundant, and I am also a big supported of the old YAGNI adage, and I think we can add a development image or a .devcontainer.json when and if someone feels like they need it.

I've had a few now, so I will make the changes in the morning to include only the production image.

scastiel:
Thanks @justcallmelarry @Max-TheCat @ChristopherJohnston for collaborating on this 😀

PR #62, opened 2024-01-25
Fix search functionality

### LINKED ISSUE #61: Search fails when enterering special characters
This issue occurs because the special characters are not properly escaped when used in a regular expression pattern.

Steps to reproduce:
1. Enter a special character like `(` or `[` in the expenses search field.
2. The application will crash.

cufarvid:
Fixes #61.

I have also added some additional information for local development setup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,10 +43,11 @@ If you want to contribute financially and help us keep the application free and 
 ## Run locally
 
 1. Clone the repository (or fork it if you intend to contribute)
-2. `npm install`
+2. Run `npm install` to install dependencies
 3. Start a PostgreSQL server. You can run `./scripts/start-local-db.sh` if you don’t have a server already.
 4. Copy the file `.env.example` as `.env`
-5. `npm run dev`
+5. Run `npm run postinstall` to apply database migrations and update Prisma Client
+6. Run `npm run dev` to start the development server
 
 ## Run in a container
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -43,10 +43,11 @@ If you want to contribute financially and help us keep the application free and
 ## Run locally
 
 1. Clone the repository (or fork it if you intend to contribute)
-2. `npm install`
+2. Run `npm install` to install dependencies
 3. Start a PostgreSQL server. You can run `./scripts/start-local-db.sh` if you don’t have a server already.
 4. Copy the file `.env.example` as `.env`
-5. `npm run dev`
+5. Run `npm run postinstall` to apply database migrations and update Prisma Client
```
scastiel, line 49: `npm run postinstall` is unnecessary as it is run by `npm install` automatically.

scastiel (approved):
Nice 👍

README.md
cufarvid: We may need to adjust the instructions or order of operations, as `postinstall` fails when the PostgreSQL server is not running.

scastiel: You’re absolutely right! Maybe put the instruction about the DB server before anything else then 😉

cufarvid: I have updated the instructions.

Issue #65, opened 2024-01-29
Store images somewhere else than S3


scastiel:
Since #63, users can attach images to expenses, but they can be stored only on S3. If someone has the need to store images somewhere else (locally, on an FTP server…), feel free to comment here with your need and we’ll see how to implement it.

mertd:
A relatively simple change could be to support S3 providers other than AWS and allow configuring the endpoint via an additional environment variable like `S_UPLOAD_ENDPOINT`.

Support for additional storage methods may not be necessary beyond this. There are FOSS options for S3 compatible storage so there is no vendor lock-in. Focusing storage options on a single method may help keeping complexity lower, too.

justcallmelarry:
Looks like next-s3-upload has support for the endpoint already:
https://next-s3-upload.codingvalue.com/bucket-config#bucket-endpoint

ChristopherJohnston:
support for storage on a local file system might be useful for homelab deployments

justcallmelarry:
> support for storage on a local file system might be useful for homelab deployments

it could, but you should also just be able to use for example [Minio](https://min.io/) in order to reach the same functionality, and since you are running multiple containers already (postgres+app) i don't think the overhead of adding for example Minio would hurt too much.
Then you can choose where to store the Minio volumes if you wish

justcallmelarry:
I tried getting Minio running locally and got it to work with uploading the image to the bucket, but not with then showing it.

I think you would have to make it work with serving images via presigned URLs (in addition to uploading via presigned URLs) in order to get it to work. I can continue looking into it another day if this would be a suitable solution?

mertd:
It seems we both gave it a try in parallel. :)

I'd be glad to hear your feedback to my attempt: #71

justcallmelarry:
> It seems we both gave it a try in parallel. 🙂
> 
> I'd be glad to hear your feedback to my attempt: #71

Looks very similar to my try! Except yours handles the next config parts way better than my solution.
It seems my issue was that I did not configure my Minio bucket correctly, so the only thing I would maybe say could be added is a hint in the documentation about Minio and how to setup the bucket. But I'm also unsure if that actually should be in the documentation for Spliit… so I did not leave it as a comment on the PR.

Very nice work!

ChristopherJohnston:
> I tried getting Minio running locally and got it to work with uploading the image to the bucket, but not with then showing it.
> 
> 
> 
[...]

PR #71, opened 2024-01-30
Support for additional S3 providers

mertd:
Adding support for additional S3 providers to avoid vendor lock-in and for a better self-hosting experience. See #65 

I successfully tested integration with a local instance of Minio. I have not tested these changes with a standard AWS S3 setup.

### DIFF
--- a/src/app/api/s3-upload/route.ts
+++ b/src/app/api/s3-upload/route.ts
@@ -1,11 +1,12 @@
 import { sanitizeKey } from 'next-s3-upload'
 import { POST as route } from 'next-s3-upload/route'
+import { env } from '@/lib/env'
 
 export const POST = route.configure({
   key(req, filename) {
     return sanitizeKey(filename).toLowerCase()
   },
-  endpoint: process.env.S3_UPLOAD_ENDPOINT,
+  endpoint: env.S3_UPLOAD_ENDPOINT,
   // forcing path style is only necessary for providers other than AWS
-  forcePathStyle: !!process.env.S3_UPLOAD_ENDPOINT,
+  forcePathStyle: !!env.S3_UPLOAD_ENDPOINT,
 })
--- a/src/app/api/s3-upload/route.ts
+++ b/src/app/api/s3-upload/route.ts
@@ -5,4 +5,11 @@ export const POST = route.configure({
   key(req, filename) {
     return sanitizeKey(filename).toLowerCase()
   },
+  accessKeyId: process.env.S3_UPLOAD_KEY,
+  secretAccessKey: process.env.S3_UPLOAD_SECRET,
+  bucket: process.env.S3_UPLOAD_BUCKET,
+  region: process.env.S3_UPLOAD_REGION,
+  endpoint: process.env.S3_UPLOAD_ENDPOINT,
+  // forcing path style is only necessary for providers other than AWS
+  forcePathStyle: !!process.env.S3_UPLOAD_ENDPOINT,
 })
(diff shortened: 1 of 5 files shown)


src/app/api/s3-upload/route.ts
```diff
@@ -5,4 +5,11 @@ export const POST = route.configure({
   key(req, filename) {
     return sanitizeKey(filename).toLowerCase()
   },
+  accessKeyId: process.env.S3_UPLOAD_KEY,
+  secretAccessKey: process.env.S3_UPLOAD_SECRET,
+  bucket: process.env.S3_UPLOAD_BUCKET,
+  region: process.env.S3_UPLOAD_REGION,
```
justcallmelarry, line 11: these are superfluous i think, only `endpoint` and `forcePathStyle` are needed if the rest are the default env vars used

src/components/expense-documents-input.tsx
```diff
@@ -29,7 +29,7 @@ type Props = {
 
 export function ExpenseDocumentsInput({ documents, updateDocuments }: Props) {
   const [pending, setPending] = useState(false)
-  const { FileInput, openFileDialog, uploadToS3 } = useS3Upload()
+  const { FileInput, openFileDialog, uploadToS3 } = usePresignedUpload() // use presigned uploads to addtionally support providers other than AWS
```
justcallmelarry, line 32: Not sure about this (since I'm no expert on js/ts), but you could continue to use the `useS3Upload()` function if `!!process.env.S3_UPLOAD_ENDPOINT`, since it's recommended for various reasons by the `next-s3-upload` lib.

So import both, and set the variables depending on the env var.

If this doesn't work for whatever client/server magic reasons that always happens with next, you may also ignore this comment :)

src/app/api/s3-upload/route.ts
mertd: Thank you! I removed them as they were not needed indeed.

src/components/expense-documents-input.tsx
justcallmelarry: Yeah, that's what I meant about the client/server magic, so then I agree that this is the better solution, but  @scastiel has the last word on the PR anyhow 😄

src/app/api/s3-upload/route.ts
```diff
@@ -5,4 +5,7 @@ export const POST = route.configure({
   key(req, filename) {
     return sanitizeKey(filename).toLowerCase()
   },
+  endpoint: process.env.S3_UPLOAD_ENDPOINT,
+  // forcing path style is only necessary for providers other than AWS
+  forcePathStyle: !!process.env.S3_UPLOAD_ENDPOINT,
```
scastiel, line 10: Prefer using `env` from `@/lib/env` when on the server, as you get type safe values 😉 

```ts
  endpoint: env.S3_UPLOAD_ENDPOINT,
  // forcing path style is only necessary for providers other than AWS
  forcePathStyle: !!env.S3_UPLOAD_ENDPOINT,
```

src/components/expense-documents-input.tsx
scastiel: Agree to use `usePresignedUpload` everytime if it doesn’t cause any issue with S3. Plus it’s tricky to use a hook conditionally with React 😉

scastiel (approved):
Amazing 💯

justcallmelarry:
I did try this out towards my own hosted version of Minio (behind a reverse proxy) and it works like a charm 👌 
I did not test it out against an actual S3 bucket, though.

src/app/api/s3-upload/route.ts
mertd: Thank you! I am now using the typed environment variables. I also resolved the merge conflicts introduced by the recent changes on `main`.

scastiel:
Just tested that everything still works with the S3 bucket 👍

PR #73, opened 2024-02-01
Add an env file with mocked env vars added for Docker production builds

justcallmelarry:
I noticed that when I was building the spliit docker image on my server, the `NEXT_PUBLIC_ENABLE_EXPENSE_DOCUMENTS` env var was not respected by my production image.

After some debugging I found that the env var is scrambled at *build time* by next, and as such must be present when building the image.

I instead added a new file (`scripts/build.env`) which is used as the normal `.env` file during docker builds.

Any new env vars would need to be added to the file for upcoming changes as well, with mocked values.


justcallmelarry:
@scastiel i know you don't really use the `Dockerfile`, but wondering if there is anything I missed? or if the PR should be discarded (I can just keep it for myself in the fork for my own workflow)

justcallmelarry:
These changes are more or less included in https://github.com/spliit-app/spliit/pull/91 so it can just be closed when and if that one is merged

justcallmelarry:
Closing due to above mention PR being merged 👆

justcallmelarry:
Closing due to above mention PR being merged 👆

PR #77, opened 2024-02-03
Update start_url to /groups page

RayBB:
This way when you install the PWA it will open to the groups page each time you open it instead of the home page.

Thanks for all the hard work on this project!


scastiel:
Makes sense 👍

Issue #72, opened 2024-02-01
Deduce category from expense title


mertd:
Automating the category selection would allow users to input and categorize their expenses faster and more reliably. Imagine a user typing their expense title, and based on their input, a category gets selected for them.

Two approaches of implementing this could be:

1. AI: Use the OpenAI integration added in #69. Prompt the latest stable turbo model for one of the available categories.
2. Keywords: For each category, maintain a set of keywords. If a keyword is found, the mapped category will be selected.

Implementing the AI option would allow categorization of many more titles than the keyword approach. It would also increase the running cost of spliit. Starting out with an implementation of keyword based mapping will be more economical. However, usability may be limited considering the range of possible titles.

I would be happy to give either of these approaches a try, with AI being my preference. Are you interested in such a feature? If yes, which approach are you leaning towards?

scastiel:
Cool idea! I’m okay with testing with OpenAI.

The cost should be pretty low for each expense, but might grow fast if we call OpenAI automatically (~50 to 100 expenses are created each day currently, and it’s growing).

What do you think of adding a subtle magic wand icon at the right of the category, and call OpenAI only when clicked? At least as a first step.

justcallmelarry:
I would like to see this as an opt-in feature, at least for self hosters (like the current OpenAI integration for pics)

To make fewer calls to OpenAI both the solutions could be used, ie (pseudo code):
```
if (any keywords in title) {
    category = hit
} else {
    category = get_category_from_openai()
}
```

In the long run:
With growing data for spliit.app, I think this is something that you could have a "buddy service" for, where you can have a model trained on the database data (you have everyones expense title and what category they gave it, while excluding all "General" categories)

scastiel:
Agree that the feature should be opt-in.

I suggest we start with a plain call to OpenAI. An optimization for later could be to persist the results to avoid doing the same call too many times. As you suggest @justcallmelarry, on the long term it could be interesting to train our own model.

mertd:
Sounds good, I will give the AI option a go then!

> ~What do you think of adding a subtle magic wand icon at the right of the category, and call OpenAI only when clicked? At least as a first step.~ Edit: let’s do that only later if the cost is too high.

Waiting until the expense title input loses focus may be a good balance between number of API calls and necessary user interactions.

> With growing data for spliit.app, I think this is something that you could have a "buddy service" for, where you can have a model trained on the database data

[...]

PR #80, opened 2024-02-04
Automatic category from expense title

mertd:
Automatically deduce category from expense title using the OpenAI API. See #72

I tried to follow the structure of the existing component and action files. Since spliit is the first project I am contributing to using Next.js and OpenAI APIs, I am particularly looking forward to feedback in this area.

The environment variable used to enable this feature could be merged with the one added in #69, e.g. `NEXT_PUBLIC_ENABLE_AI`. Similar to the receipt extraction feature, we could also add a "Beta" note to the category field. Please let me know your preferences.

### DIFF
--- a/src/app/groups/[groupId]/expenses/create-from-receipt-button-actions.ts
+++ b/src/app/groups/[groupId]/expenses/create-from-receipt-button-actions.ts
@@ -23,7 +23,7 @@ export async function extractExpenseInformationFromImage(imageUrl: string) {
               This image contains a receipt.
               Read the total amount and store it as a non-formatted number without any other text or currency.
               Then guess the category for this receipt amoung the following categories and store its ID: ${categories.map(
-                (category) => formatCategory(category)
+                (category) => formatCategory(category),
               )}.
               Guess the expense’s date and store it as yyyy-mm-dd.
               Guess a title for the expense.
--- a/src/app/groups/[groupId]/expenses/create-from-receipt-button-actions.ts
+++ b/src/app/groups/[groupId]/expenses/create-from-receipt-button-actions.ts
@@ -2,6 +2,7 @@
 import { getCategories } from '@/lib/api'
 import { env } from '@/lib/env'
 import OpenAI from 'openai'
+import { ChatCompletionCreateParamsNonStreaming } from 'openai/resources/index.mjs'
 
 const openai = new OpenAI({ apiKey: env.OPENAI_API_KEY })
 
@@ -9,7 +10,7 @@ export async function extractExpenseInformationFromImage(imageUrl: string) {
   'use server'
   const categories = await getCategories()
 
-  const body = {
+  const body: ChatCompletionCreateParamsNonStreaming = {
     model: 'gpt-4-vision-preview',
     messages: [
       {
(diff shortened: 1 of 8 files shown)


mertd:
I noticed that the receipt extract feature was not yet using presigned uploads introduced in #65. I updated this and added a note to the readme that S3 endpoints need to be public.

src/app/groups/[groupId]/expenses/create-from-receipt-button-actions.ts
```diff
@@ -1,15 +1,17 @@
...
+import { ChatCompletionCreateParamsNonStreaming } from 'openai/resources/index.mjs'
 
 const openai = new OpenAI({ apiKey: env.OPENAI_API_KEY })
 
 export async function extractExpenseInformationFromImage(imageUrl: string) {
   'use server'
   const categories = await getCategories()
 
-  const body = {
+  const body: ChatCompletionCreateParamsNonStreaming = {
```
Comment (scastiel) [inline src/app/groups/[groupId]/expenses/create-from-receipt-button-actions.ts:14]: 💯

src/app/groups/[groupId]/expenses/create-from-receipt-button.tsx
```diff
@@ -46,7 +46,7 @@ export function CreateFromReceiptButton({
   categories,
 }: Props) {
   const [pending, setPending] = useState(false)
-  const { uploadToS3, FileInput, openFileDialog } = useS3Upload()
+  const { uploadToS3, FileInput, openFileDialog } = usePresignedUpload()
```
Comment (scastiel) [inline src/app/groups/[groupId]/expenses/create-from-receipt-button.tsx:49]: 👍

src/lib/utils.ts
```diff
@@ -15,3 +16,8 @@ export function formatExpenseDate(date: Date) {
     timeZone: 'UTC',
   })
 }
+
+/** Format category, e.g. for use in AI prompts */
+export function formatCategory(category: Category) {
```
scastiel, line 21: I suggest the function name to be more specific, e.g. `formatCategoryForAIPrompt`. Otherwise, one could think this should be used to display a category on the UI.

src/components/expense-form-actions.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { getCategories } from '@/lib/api'
+import { env } from '@/lib/env'
+import { formatCategory } from '@/lib/utils'
+import OpenAI from 'openai'
+import { ChatCompletionCreateParamsNonStreaming } from 'openai/resources/index.mjs'
+
+const openai = new OpenAI({ apiKey: env.OPENAI_API_KEY })
+
+/** Limit of characters to be evaluated. May help avoiding abuse when using AI. */
+const limit = 40 // ~10 tokens
```
scastiel, line 11: Good idea!

scastiel (approved):
Amazing, it works so well!

An improvement suggestion (not a blocker for the PR): when the user enters a title and leaves the input, we could display a tiny loader at the right of the field to indicate there is something pending, until we get the result. @mertd let me know if you want to add it before I merge, if you want to do it later, or not 😉

mertd:
Thank you! What do you think of replacing the dropdown arrow like this when loading?
image

scastiel:
> Thank you! I like your suggestion. What do you think of replacing the dropdown arrow with a spinner like this when loading? image

That’d be perfect!

src/lib/utils.ts
mertd: Thanks! I renamed it and removed the TSDoc since I felt it was redundant with the more concise name.

PR #81, opened 2024-02-04
Format currency with thousand separators

justcallmelarry:
Added a formatting function for currency strings, in order to be able to use (and change) the format everywhere at once, and alos to add thousand separators to the formats.

I chose to do this by using the `en-US` locale (which is also used for dates), and left it hard coded for now. But since I've seen issues about translations etc, I thought perhaps a group-wide locale setting would be implemented at some point, and this makes it somewhat easy to build upon.

This makes it easier to parse larger numbers especially in the list, but also in the balances overview. This is less needed for currencies like USD and EUR, but makes more sense for other currencies like (my use case) SEK, or something like JPY.


scastiel:
Nice 💯

Issue #85, opened 2024-02-05
Wrong / unfair rounding


lfilho:
Thanks for the great project, appreciate it! Just used it for testing and found that a split was not very precise/fair, see screenshots:

![CleanShot 2024-02-05 at 13 58 54](https://github.com/spliit-app/spliit/assets/282756/209b7df4-7083-4899-a2ac-49d73ef860b9)

![CleanShot 2024-02-05 at 13 59 40](https://github.com/spliit-app/spliit/assets/282756/d80357a6-c049-4e69-8016-ef88d9ae6a47)

scastiel:
Thanks for reporting this!

Here is the reason of this result:
- Luiz paid $5 for 3 people including himself, so his balance is $3.34 (rounded up)
- Luiz paid $10 for 3 people including himself, so his balance increases of $6.67 (rounded up), to $10.01

The question here is: should we round up the amounts for the payer? Or round down? Or round to the closest? I understand that in this situation it might seem weird, but trust me, none of these solutions will give the intuitively expected result every time.

Maybe it is possible to round balances later (instead of rounding it for each transaction) or find a more complex algorithm. I’m open to suggestions 😉

justcallmelarry:
In my opinion and experience (I used to work for a fintech company) doing a expense by expense rounding can lead to huge inconsistencies.

The above example is a pretty good example if taken to an extreme. If you add a hundred $10 expenses which the same person pays (let's say it's a weekly cost someone pays for every week) this will end up being an entire dollar in "rounding errors".

### Test similar to above example
I did some testing on doing only rounding on the totals in a [branch in my fork](https://github.com/justcallmelarry/spliit/tree/feature/round-totals-only) found that this solves the current issue:
<img width="776" alt="Screenshot 2024-02-07 at 16 52 54" src="https://github.com/spliit-app/spliit/assets/24991021/b3d58094-d85f-4686-bd26-0086c0b954ee">
<img width="769" alt="Screenshot 2024-02-07 at 16 53 01" src="https://github.com/spliit-app/spliit/assets/24991021/39fd190c-cbd9-48a9-b312-265010c81120">

### Test with more added data
But causes issues when another $10 expense is added:
<img width="764" alt="Screenshot 2024-02-07 at 16 53 25" src="https://github.com/spliit-app/spliit/assets/24991021/ddaa0ad3-d95f-47ab-9dce-567eb5e2d38a">
[...]

PR #88, opened 2024-02-09
Round totals rather than expense by expense

justcallmelarry:
## Problem statement
As raised in https://github.com/spliit-app/spliit/issues/85, balances get a bit weird when they are not equally divisible between all participants.

Current logic does rounding on an expense by expense basis, which ends up with some inconsistencies. I made three groups with three users each, which I then added ten $10 expenses to, and wound up with the following results:
<details>
  <summary>Images</summary>
  Screenshot 2024-02-09 at 16 19 59  
  Screenshot 2024-02-09 at 16 21 34
  Screenshot 2024-02-09 at 16 22 59
</details>

So it's not only about who pays for expenses, but also the order of the users has an impact on who pays what.
This seems inherently incorrect/unfair, so I set out to try to fix the calculations.

### Proposed solution 1
Since there is nothing wrong with the general setup of calculating the balances, I just tried to change the rounding parts and leave the rest as is.
Do not do any rounding when calculating balances, instead only do rounding once the totals for each user has been calculated.

### Issues with the solution
I noticed that sometimes the balances got offset by `0.01` (`+16.67` / `-16.66`), which looked ugly (and perhaps confusing). Furthermore, when all the reimbursements were done, the extra `0.01` was still left as a positive balance, even though no reimbursements were listed.

### Proposed solution 2
Instead of using the balances as they are, use them to calculate the reimbursements just like before, but then construct something I called "public balances" from those reimbursements, making sure that the balances will always show the same amounts as the reimbursements

### DIFF
--- a/src/app/groups/[groupId]/balances/page.tsx
+++ b/src/app/groups/[groupId]/balances/page.tsx
@@ -9,7 +9,11 @@ import {
   CardTitle,
 } from '@/components/ui/card'
 import { getGroupExpenses } from '@/lib/api'
-import { getBalances, getSuggestedReimbursements } from '@/lib/balances'
+import {
+  getBalances,
+  getPublicBalances,
+  getSuggestedReimbursements,
+} from '@/lib/balances'
 import { Metadata } from 'next'
 import { notFound } from 'next/navigation'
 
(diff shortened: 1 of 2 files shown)


scastiel (approved):
I love your changes in the algorithm. To be honest I can’t promise we’ll never find some cases where the result isn’t what we expect, but it’s definitely worth trying :)

To answer your concern, I can’t image groups with so many participants that scaling it becomes an issue ;)

PR #89, opened 2024-02-10
Add Support for mySQL / mariaDB Database

shynst:
## Motivation

I wanted to use vercel to publish the app while still being able to keep the database stored on my server. However, since my host sadly doesn't support postgres, I decided to **add mySQL support** to spliit. As an added bonus, the app now runs a lot snappier! Somehow postgres database queries on my vercel-free-tier where unbearably slow (on average somewhere between 500 and 1000ms).

I can imagine that others would be interested in using mySQL, too. That's why I created this pull request.

## Implementation

Since spliit is using prisma as ORM, it should be quite straightforward to apply the necessary changes.

My first idea was to add an environment variable to `prisma/schema.prisma` to configure the provider like this

```
datasource db {
  provider = env("PRISMA_PROVIDER")
  ...
}
```

However, this doesn't work (providers cannot be set via environment variables)!

Therefore I decided to generate `prisma/schema.prisma` from a template file `prisma/schema.prisma.template` with a script that is invoked with `package.json`'s postinstall hook (`scripts/generate-prisma-schema.sh`). The provider to be used is determined from the database URL scheme provided in `.env` environment configuration. If the URL starts with `postgresql://`we are using postgres, if it starts with `mysql://`we are using mySQL, ...

Almost there ...

Migrations are platform specify, too. Therefore the migrations specified in `prisma/migrations` will fail for mySQL.
Therefore, I had to

- move existing `prisma/migrations` to `prisma/migrations_postgresql`
- add new migrations for mySQL in `prisma/migrations_mysql`
- extend the script `scripts/generate-prisma-schema.sh` to symlink `prisma/migrations` to the correct migration folder

## Drawbacks

- needs post-install hook to work
- migrations have to be specified for each database platform separately (maybe someone could point me to a more elegant solution)

## Advantages

- works for mySQL and mariaDB
- easily extendable to other database platforms


shynst:
Probably this is useful only to myself. If so please tell me and I will close this PR.

ChristopherJohnston:
I think it is useful to give those who are self-hosting the choice of database engine.

As your two open PRs are quite big, it may be that @scastiel hasn't yet had time to look at them fully so keep it around for now and I'm sure you'll get some feedback.

PR #90, opened 2024-02-10
Fix decimal point issue in expense form schema `amount` field

### LINKED ISSUE #78: Decimal separator for locales where decimal point is a comma
> Unfortunately, the "amount" entry field only allows a dot ( . ) as decimal separator, not a comma ( , ). This is not a problem on desktop computers with a keyboard, however on iPhone with german keyboard layout, I am presented with a keyboard containing 0-9 and the comma only (due to the input field type I guess). I have no possibility to enter a dot instead of a comma, and therefore cannot add an expense. I hope you're able to fix that, that would be awesome!

jantuomi:
### Problem

The `amount` input field in the "add expense" view uses the `decimal` inputMode. This renders a numbers-only keyboard that on some locales only allows the user to input a comma as the decimal point. However, the zod schema does not permit a comma as the decimal point, since `Number(value)` will return `NaN` for such numbers.

E.g. `Number("1.23") === 1.23`, but `Number("1,23") === NaN`.

### Solution

Replace commas with dots in the user-inputted amount value before casting it to a JS number. The rest of the schema validation will proceed as normal.

Fixes #78

### DIFF
--- a/src/lib/schemas.ts
+++ b/src/lib/schemas.ts
@@ -51,7 +51,8 @@ export const expenseFormSchema = z
         [
           z.number(),
           z.string().transform((value, ctx) => {
-            const valueAsNumber = Number(value)
+            const normalizedValue = value.replace(/,/g, '.')
+            const valueAsNumber = Number(normalizedValue)
             if (Number.isNaN(valueAsNumber))
               ctx.addIssue({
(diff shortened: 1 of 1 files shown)


scastiel (approved):
👍 🥳

Issue #82, opened 2024-02-05
Add charts by category


scastiel:
Not that each expense can be assigned a category, it would be cool to add a new tab for groups to display a chart of total amount by category. Ideally, there would be a date range selector (see [component on shadcn UI](https://ui.shadcn.com/docs/components/date-picker#date-range-picker)), but doesn’t have to be here in the first version.

I’m not fully aware of the best choice for a chart library for React, open to suggestions :)

justcallmelarry:
Another chart I thought of is a latest X months stacked area chart with different values for the different categories as stacks.

I've used [Recharts](https://recharts.org/en-US/) in a hobby project, but found the styling to be tricky (but that is most likely a developer issue in my case)

I thought I'd mention it as something to look into, at least. Also asked a friend who is a frontender what he would use if he was to choose something new today.

alessiozanutta:
is there a way to see something more under stats?
i really would like to see some charts (i think it's the final purpose of process of keeping track of all the expenses) to realize and be aware of where goes my money.
without this "analysis" i have to export all the entries and build charts with other programs.
thinking that this should really have the priority of you wonderful guys that already came up with this very good (and free for the folks) alternative to splitwise.

PR #94, opened 2024-02-18
feat: initialise a new totals tab with basic UI

### LINKED ISSUE #93: Feature request: group spending tab similar to splitwise
I have written my own script to calculate this from the json dump that spliit generates. But would be very useful if the same is shown on the UI itself
ref:

meta-boy:
A new tab named 'Stats' with a simple UI that shows the group-level spending summary. Can be further extended to show other stats, charts, etc.
closes #93 

Preview:

image


justcallmelarry:
I don't have much to contribute here, but I thought I would suggest using "Stats" (or similar) instead of "Totals" as you could then expand it with other things, like spending charts etc raised in https://github.com/spliit-app/spliit/issues/82

meta-boy:
Oh I agree. That would be much better. I was just making it closer to what splitwise has to offer but definitely could go with "Stats" and further add spending charts to it like you mentioned

justcallmelarry:
I'm also wondering if this takes into account not having a default active user set? It looks like it's still going through all of the expenses when it could just exit the functions in case the active user is `''` without doing any calculations.
But will all the boxes still render? imo `Total you paid for` and `Your total share` should not show if you do not have an active user set.

meta-boy:
> I'm also wondering if this takes into account not having a default active user set? It looks like it's still going through all of the expenses when it could just exit the functions in case the active user is `''` without doing any calculations (this is an improvement I would suggest in either case).
> But will all the boxes still render? imo `Total you paid for` and `Your total share` should not show if you do not have an active user set.

Agreed, will make the change

src/app/groups/[groupId]/stats/page.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const expenses = await getGroupExpenses(groupId)
+  const totalGroupSpendings = getTotalGroupSpending(expenses)
+
+  return (
+    <>
+      <Card className="mb-4">
+        <CardHeader>
+          <CardTitle>Totals</CardTitle>
+          <CardDescription>
+          Spending summary of the entire group
```
Comment (scastiel) [inline src/app/groups/[groupId]/stats/page.tsx:39]:
```suggestion
          Spending summary of the entire group.
```

scastiel (approved):
That’s great, thanks for the contribution! We’ll be able to use the same tab to offer stats by category 😉

Issue #30, opened 2024-01-08
Paginate expenses


scastiel:
The expense list can be slow when displaying several hundreds of transactions. It would make sense to paginate transactions, i.e. either:

1. Display the first X transactions by default, and then a “Show more” button
2. Display an actual pagination, i.e. show page one by default and offer the option to go to a random page.

My preference is on option 1, but open to discussion.

justcallmelarry:
I'll add that i like option 1 as well. As long as you can search and export I don't think you really need option 2.

manuerwin:
Agree, option 1 would work for me so long as search works well

shynst:
I would prefer a technique called "infinite scroll" over traditional "old-school" pagination.

**Infinite Scroll** is a technique that automatically adds the next page as the user scrolls down through content, more content is loaded. User don’t need to wait for pages to preload. So infinite scroll can provide user a better experience.

https://blog.logrocket.com/3-ways-implement-infinite-scroll-react/

manuerwin:
> I would prefer a technique called "infinite scroll" over traditional "old-school" pagination.
> Infinite Scroll is a technique that automatically adds the next page as the user scrolls down through content, more content is loaded. User don’t need to wait for pages to preload. So infinite scroll can provide user a better experience.

I change my vote to this "Option 3" :)

shynst:
Here is my implementation of "option 3: infinite scroll": #95

ChristopherJohnston:
@scastiel I think this can now be closed after the #95 merge

Issue #22, opened 2024-01-06
Import from Splitwise


rgov:
Splitwise offers an "Export as spreadsheet" feature. It would be useful to be able to import this data into Spliit for continuity of groups that were created before.

Splitwise exports CSV files with the following headers:

|Date|Description|Category|Cost|Currency|UserA|UserB|...|
|---|---|---|---|---|---|---|---|

Values in the User columns are positive for users who paid and negative for users who owed. For example, a $100 purchase that was paid for by UserA and split evenly will be listed as

|Date|Description|Category|Cost|Currency|UserA|UserB|
|---|---|---|---|---|---|---|
|2024-01-01|Something|General|300.00|USD|150.00|-150.00|

The final row is the total balance per user, which is the column sum for that user. There is no need to import this, but for the record, a positive value represents that the user "gets back" some amount, whereas a negative value represents that the user "owes" the amount. 

This would require #6, at least the ability to split unevenly by dollar amount.

I'd suggest matching users by column name and error if a match isn't found; it's very easy to edit the CSV file if the name needs to be changed, no need to create new UI for it.

ChristopherJohnston:
I have the beginnings of an implementation as a script here: https://github.com/ChristopherJohnston/spliit2/blob/import-splitwise/src/scripts/import.ts

This was created before the recent uneven split addition, and also assumes implementation of expenseDate and categories per my recent PRs....I'll update it accordingly once these are done.

I was able to match balances on a split wise group that has ~2800 lines in the export (granted this is a group of 2 participants where all of the expenses are equal split)

This script would require hosting your own instance of the app but could probably be used as part of a future "create group from csv" implementation

scastiel:
I love this! It shouldn’t be too hard to develop a MVP of an import feature (e.g. assuming that the user names match exactly). @ChristopherJohnston tell me if you’re interested to work on such a feature, otherwise I’ll take care of it 😃

ChristopherJohnston:
> @ChristopherJohnston tell me if you’re interested to work on such a feature, otherwise I’ll take care of it 😃

I'm not particularly skilled at the UI part, so the file upload part would probably slow me down somewhat!

ChristopherJohnston:
The export is essentially a ledger so some of the information about how the expense was split is lost. It's possible to infer but not sure if this is always going to be correct.

A few hurdles I've seen so far:

[...]

PR #95, opened 2024-02-18
Implement "infinite scroll" for expenses

shynst:
## Motivation

As @scastiel mentioned in #30 the expense list gets quite slow when dealing with a large number of transactions. I have managed to import my "home expenses" group over from Splitwise which dates back to 2018 and features around 3000 expenses. To convert my expenses from Splitwise I used the [script from @ChristopherJohnston](https://github.com/spliit-app/spliit/issues/22#issuecomment-1881632719) as starting point.

Whenever I was entering this group I needed to wait for about 15 seconds until the entries appeared. When it was fully loaded the UI felt a bit sluggish - buttons took some time to respond to clicks.

Therefore I was motivated to implement some sort of incremental expense loading. As it is 2024, I went for a technique called "infinite scroll" over traditional "old-school" pagination. **Infinite Scroll** is a technique that automatically adds the next page as the user scrolls down through content, more content is loaded. User don’t need to wait for pages to preload. So infinite scroll can provide user a better experience.

## Implementation

Changes to component `Page`

  - only load the first 200 expenses (if there are 200 or more)
  - pass preloaded expenses and total expense count to child component `ExpenseList`

Changes to component `ExpenseList`

- Extract `ExpenseCard` (the component that renders an expense list item) from `ExpenseList`
- If there are more expenses to show:
  - append preloading "skeletons" to end of list
  - fetch more expenses when last item in list comes into view (using [`react-intersection-observer`](https://github.com/researchgate/react-intersection-observer))
  - after each fetch increase number of fetched expenses by 1.5 (*)

(*) Database fetch is usually not the issue here, but the length of the list is. The longer the list gets, the longer React needs to redraw. To limit the number of redraws, the next query fetches 1.5 times more items. Therefore, when the list is still short, we add less items (which is fast). The longer the list gets, the more items are added. This is slower, but the list will be completed with less redraws (which is less painful).

## Links

Some resources that helped me implementing infinite scroll

- https://blog.logrocket.com/3-ways-implement-infinite-scroll-react/
- https://medium.com/@ryanmambou/how-to-implement-infinite-scroll-in-nextjs-without-button-99d8ce886985
- https://github.com/researchgate/react-intersection-observer

## Remarks

- dramatic speedup for long lists (startup time is down from 15 to 2 secs)
- for 99% we don't need to load more that the most recent 200 entries
- UI feels a lot more responsive, too
- no "load more" button or "old-school" pagination needed
- works very well with the search button, too

Enjoy! Hope you like it!
Comments and suggestions are always welcome.


ChristopherJohnston:
This is a great contribution! I also have a split wise-imported expense list (via a script) with similar issues as a result of its size.

nb would also be good to share how you imported - my script is here: https://gist.github.com/ChristopherJohnston/a6e69bd8894f20ecfe127fa4149bd013

shynst:
@scastiel : I think this is a good contribution to spliit. Did you have time to look over it? Looking forward to your input.

ChristopherJohnston:
@scastiel the app is definitely in need of something like this. It is very slow to render the expense list when there is a reasonably large number of expense items (ie 1000+)

ChristopherJohnston:
@shynst are the merge conflicts simple to resolve? This might help things along

shynst:
@ChristopherJohnston should be ready to merge now!

ChristopherJohnston:
Thanks @shynst! @scastiel - when do you think you will be able to catch up on these PRs?

scastiel:
Amazing @shynst!

PR #104, opened 2024-02-20
Add onClick-event to select all to amount input

justcallmelarry:
When you add a new expense you always have to start by removing the placeholder 0 value that is already there.

I added an onFocus event to select the entire value so that you can just start writing the amount immediately

### DIFF
--- a/src/components/expense-form.tsx
+++ b/src/components/expense-form.tsx
@@ -222,6 +222,7 @@ export function ExpenseForm({
                         inputMode="decimal"
                         step={0.01}
                         placeholder="0.00"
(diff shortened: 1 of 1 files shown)


justcallmelarry:
> Works fine with Chrome, has issues though with Safari when selecting the input field by clicking into it:
> 
> * first the onFocus event fires, selecting the text
> * after that the onClick event fires, **deselecting** it again
> 
> ### Suggestion
> Use an `onClick()` event handler to select the text instead.

Good catch, I had only tested in safari on mobile and there it worked as expected. I think onClick might be good.

scastiel:
💯

PR #105, opened 2024-02-20
Add a prettier script for ease of use (npm run prettier)

justcallmelarry:
I'm not sure if I have things installed incorrectly or something on my machine, but I can't seem to run prettier normally, so I added a script for it locally so that I can just run `npm run prettier` in order to fix formatting.

I thought I'd share, but feel free to shoot down this PR if this isn't something you want in the repo.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
     "lint": "next lint",
     "check-types": "tsc --noEmit",
     "check-formatting": "prettier -c src",
+    "prettier": "prettier -w src",
     "postinstall": "prisma migrate deploy && prisma generate",
     "build-image": "./scripts/build-image.sh",
     "start-container": "docker compose --env-file container.env up"


shynst:
Which IDE / editor are you using?

I am using Visual Studio Code and I like prettier rules to be applied whenever I save a file. So there is no need to remember having to run a script to get formatting right.

If you are using VSCode too, just add this lines to `.vscode/settings.json`

```js
{
  "[javascript]": {
    "editor.formatOnSave": true
  },
  "[typescript]": {
    "editor.formatOnSave": true
  },
  "[typescriptreact]": {
    "editor.formatOnSave": true
  },
  "[json]": {
    "editor.formatOnSave": true
  },
  "[jsonc]": {
    "editor.formatOnSave": true
  },
  ...
}
``

justcallmelarry:
> Which IDE / editor are you using?
> 
> I am using Visual Studio Code and I like prettier rules to be applied whenever I save a file. So there is no need to remember having to run a script to get formatting right.
> 
> If you are using VSCode too, just add these lines to `.vscode/settings.json`
> 
> ```js
> {
>   "[javascript]": {
>     "editor.formatOnSave": true
>   },
>   "[typescript]": {
>     "editor.formatOnSave": true
>   },
>   "[typescriptreact]": {
>     "editor.formatOnSave": true
>   },
>   "[json]": {
>     "editor.formatOnSave": true
>   },
>   "[jsonc]": {
>     "editor.formatOnSave": true
>   },
>   ...
> }
> ```

I do use VSCode generally (and sometimes Sublime) but I prefer not to have these things set globally at least, and I generally don't use format on save as I don't like the code "jumping around" as I type it, as I'm a compulsive saver…

So I prefer to do changes -> see that they work -> format the code -> commit&push as my workflow.

It's nice to have these settings pointed out to others though! Does these settings respect the chosen ESLint/prettier settings for the project even if I have the repo open as a subdir in another project rather than opening the project on it's own in VSCode?

shynst:
I was like you. Initially I hated tools like prettier messing around with my formatting. After I while I gave up - and now I love it and can't live without it! I write my code without any formatting rules at all, then hit save and watch the code unwrapping. :)

The file `.vscode\settings` is to be placed under project root, so everything you put in there is a local workspace setting.

> Does these settings respect the chosen ESLint/prettier settings for the project even if I have the repo open as a subdir in another project rather than opening the project on it's own in VSCode?

You have to try, I don't know honestly.

PR #107, opened 2024-02-21
Add "x" button to cancel search in search bar

shynst:
Simply adds an "x" button to the search bar that will show when text has been entered.
As expected, clicking on "x" will clear the search.

Motivation for this PR: I have a long list. Having to click backspace multiple times to remove the search is time consuming, since every keystroke (backspace too) triggers a new search.


ChristopherJohnston:
this makes sense to me.

webzit:
Thank you, that is a good addition. (At the same time it would be useful if text entered in the search bar would be preserved when adding a new entry).

ChristopherJohnston:
I'm still very supportive of this PR!

scastiel:
Thanks @shynst!

PR #113, opened 2024-02-25
Fix the default value for the expense shares field

### LINKED ISSUE #112: [Bug] Default share value is 0.01 (sent as 1), should be 1 (sent as 100)
If you create an expense with a split mode 'Unevenly - By shares' and leave some share value at default, it will use 0.01 although 1 is displayed. If you edit share values, they are correctly multiplied by 100.

sashkent3:
Closes #112

### DIFF
--- a/src/components/expense-form.tsx
+++ b/src/components/expense-form.tsx
@@ -122,7 +122,7 @@ export function ExpenseForm({
           // paid for all, split evenly
           paidFor: group.participants.map(({ id }) => ({
             participant: id,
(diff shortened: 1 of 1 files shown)


shynst (approved):
Works!

scastiel:
💯

sashkent3:
@scastiel Done, my bad!

scastiel:
> @scastiel Done, my bad!

No worries, thanks 😊

PR #114, opened 2024-02-25
feature: recurring transactions added- migrations, ui, api updated. Closes #5

### LINKED ISSUE #5: recurring expenses?
hi!
really loving and appreciating this project.
i was wondering whether recurring expenses are a feature that you'd consider adding.
with that, transitioning from splitwise would be a no-brainer, for me and i'm assuming for others.
thanks so much

neonshobhit:
This Pull requests solves recurring transactions adding feature.
Closes #5


neonshobhit:
If recurring transaction is selected from the drop down menu while adding an expense, that expense id will be inserted in the recurring transaction table once.
When listing for the groups, it will check if any pending recursive transaction needed to be added, if yes, then same transaction will be copied.

WIP:
- Need to add same support while editing the expense.
- Delete only latest transaction in the recursive transaction is yet to be supported. This will be done by giving options for archiving the expense/ delete this and all future expenses.
- Need to write cron which will keep on checking any group has recursive transaction pending (without user needing to list the group expenses)

neonshobhit:
Features:
- Set the recurring expense schedule
- While fetching Group's Expenses, checks if any pending expense exists. If yes, adds into the DB and then lists all the expenses.
- Edit the expense schedule.
- Delete the expense schedule (Deleting the last added expense deletes the future entries.)
- Archive the expense which needs to be shown in history but not needed while calculating.
- Archiving the last expense doesn't add to calculations, but will occur in the next cycle.


Screenshot 2024-03-08 at 4 49 22 PM
Screenshot 2024-03-08 at 4 49 47 PM
Screenshot 2024-03-08 at 4 49 59 PM
Screenshot 2024-03-08 at 4 50 11 PM

src/app/groups/[groupId]/expenses/page.tsx
```diff
@@ -100,7 +100,7 @@ export default async function GroupExpensesPage({
 async function Expenses({ groupId }: { groupId: string }) {
   const group = await cached.getGroup(groupId)
   if (!group) notFound()
-  const expenses = await getGroupExpenses(group.id)
+  const expenses = (await getGroupExpenses(group.id))// .filter(e => !e.isArchive)
```
Comment (rgov) [inline src/app/groups/[groupId]/expenses/page.tsx:103]: Can you revert this no-op change?

package.json
```diff
@@ -54,13 +55,12 @@
...
     "vaul": "^0.8.0",
-    "zod": "^3.22.4",
-    "prisma": "^5.7.0"
+    "zod": "^3.22.4"
   },
   "devDependencies": {
     "@total-typescript/ts-reset": "^0.5.1",
     "@types/content-disposition": "^0.5.8",
-    "@types/node": "^20",
+    "@types/node": "^20.11.20",
```
rgov, line 63: This file (and presumably `package-lock.json`) is virtually unchanged except for sorting the packages and bumping the `@types/node` version. Is the version bump necessary? If not, I recommend removing both `package.json` and `package-lock.json` from this PR.

src/app/groups/[groupId]/expenses/[expenseId]/edit/page.tsx
```diff
@@ -36,7 +37,13 @@ export default async function EditExpensePage({
 
   async function deleteExpenseAction() {
     'use server'
-    await deleteExpense(expenseId)
+    await deleteExpense(groupId, expenseId)
+    redirect(`/groups/${groupId}`)
+  }
+
+  async function archiveExpenseStateUpdateAction(state: boolean) {
```
Comment (rgov) [inline src/app/groups/[groupId]/expenses/[expenseId]/edit/page.tsx:44]: Is this stuff about archiving related to this commit "added locks to ensure only 1 user can get the group expenses created"? If no, can it be factored out of this commit / pull request and submitted separately?

src/lib/api.ts
```diff
@@ -27,6 +35,134 @@ export async function createGroup(groupFormValues: GroupFormValues) {
   })
 }
 
+export async function acquireLockRecurringTransaction(groupId:string, expenseId:string, lockId: string): Promise<RecurringTransactions[]> {
```
rgov, line 38: Could you elaborate on why we need to implement locking and retrying here? What is the issue if multiple users edit a recurring expense concurrently? Some comments would go a long way towards explaining the logic.

src/lib/api.ts
```diff
@@ -27,6 +35,134 @@ export async function createGroup(groupFormValues: GroupFormValues) {
...
+    if (!existingLock.length) {
+      const query = `
+        INSERT INTO "RecurringTransactions"
+          ("groupId", "expenseId", "createNextAt", "lockId")
+          VALUES
+          ('${groupId}', '${expenseId}', 0, '${lockId}')
+          ON CONFLICT("groupId", "expenseId")
+          DO NOTHING;
+      `
+      await prisma.$queryRawUnsafe(query)
```
rgov, line 58: I don't know Prisma but I do not recommend building SQL queries via string interpolation. This is how SQL injection vulnerabilities arise, not the "unsafe" in the function name.

rgov (changes requested):

src/app/groups/[groupId]/expenses/[expenseId]/edit/page.tsx
Comment (neonshobhit) [reply src/app/groups/[groupId]/expenses/[expenseId]/edit/page.tsx:44]: No
This feature was added here for removing a recurring transction temporarily.
For example:
I have a monthly subscription of Netflix. It was there for 6 months but we cancelled for 1 month.
It will automatically add in the recurring expense section.
If I delete the transaction, future recurring transactions will not be created. Hence Archiving a transaction will result it to be there, but won't add in the split calculations.

src/lib/api.ts
neonshobhit: When someone fetches group expenses, only then I'm checking if there's a pending recurring transaction. If yes, then I insert in the DB as new transaction.

Locking is necessary to avoid race condition:
if two people fetch expenses at the same time, APIs of both will result in fetching of pending transactions and inserting in DB. Hence Double insertions for the pending transaction.

Same condition will be there for update. One person is updating the transaction and other is inserting in DB.

Locking ensures consistency and correctness.


Retrying is there to retry acquiring locks after some time if it was acquired by someone else at that instant.

src/lib/api.ts
neonshobhit: Agreed. I searched for Prisma docs but there's no option for 'createIfNotExists' hence used this method. I need this functionality for avoiding race conditions in between of inserting and fetching.

I also don't like building SQL queries like this - can you please help here for the alternatives for the above requirement?

src/app/groups/[groupId]/expenses/[expenseId]/edit/page.tsx
Comment (neonshobhit) [reply src/app/groups/[groupId]/expenses/[expenseId]/edit/page.tsx:44]: Please check [this image](https://private-user-images.githubusercontent.com/52707760/311215459-19c44392-7393-4045-b62b-454f29e3f125.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTI3ODAxMDksIm5iZiI6MTcxMjc3OTgwOSwicGF0aCI6Ii81MjcwNzc2MC8zMTEyMTU0NTktMTljNDQzOTItNzM5My00MDQ1LWI2MmItNDU0ZjI5ZTNmMTI1LnBuZz9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA0MTAlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwNDEwVDIwMTAwOVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWRlOGFhODkyOWU3Y2YzZTgwZGE3YzE4MjNlYWVmZGI2MTRlM2RkMjY2MDM2YjU0OWM2YTEwYWFhMGJhYmFjNjAmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.NR9C9_DO80OMkDkzg6frjjxQTT2dpFbJVssUq4ya8Hg) for archived expense in the fetched expenses

Comment (rgov) [reply src/app/groups/[groupId]/expenses/[expenseId]/edit/page.tsx:44]: I see. I recommend simplifying the pull request to remove archived expenses for now, and propose that as a separate feature. There may be disagreement over the user experience here and it will delay acceptance of the recurring expense feature. (The word "archive" to me does not connote the behavior of the feature, for example.)

Splitwise behavior is: The most recent expense is the one marked "recurring." Older expenses can be deleted without affecting the recurrence. If you delete the most recent expense, the recurrence stops.

That makes it easy to delete transactions 1...n-1 without affecting the recurrence.But I agree it is useful to have a way to delete the most recent transaction without affecting the recurrence, too.

src/lib/api.ts
rgov: Is it possible to delegate this concern to the database? Suppose each expense ID in a recurrence followed the pattern `ffffff01`, `ffffff02`, `ffffff03`, .... When expenses are fetched, you determine the need to create expense `ffffff04`. You do some kind of `INSERT IF NOT EXISTS` query. If there is a race and this expense already got inserted by another request, it doesn't really affect anything.

Worst case scenario I think is if, before your request completes, the expense has already been created *and* then subsequently updated. You might have a stale view of the totals. But it would only be a temporary issue.

src/app/groups/[groupId]/expenses/[expenseId]/edit/page.tsx
Comment (neonshobhit) [reply src/app/groups/[groupId]/expenses/[expenseId]/edit/page.tsx:44]: sure. Can do. I'll raise that feature as another PR and add an alert.
Archive seemed best to me as it just serves as a historical preserved record for reference. Any other suggestion?

src/lib/api.ts
neonshobhit: Umm, not.
Let's take an example:
- 1 Jan 2024 (monthly txn) `ffffff01` was added. And the group wasn't opened until now.
If Person P1 and P2 both open it at the same moment - both will determine that they need to insert 3 new Txns. How will I verify if P1 is already adding the transaction and P2 doesn't need to?
`ffffff04` can be a copy of `ffffff02` (determined by P1) or `ffffff02` (determined by P2).
For realising that someone else is working on this resource, I need to insert a reference id in DB.
That Id can be in the same Expense table, or a new table that I created.
New table made it easy to manage it as I can keep track which txn needs to be inserted without looking into Expense table + separation of concern.

So if lock is not there, in worst case both will insert the same records.

Another case:
P1 didn't call fetch groups (consider he didn't close tab for 4 months). P2 fetches now and he realizes he has to insert 3 new records.

P1 edits the amount from 100 to 200.
What shall happen now? Either P1 should successfully edit first and P2 shall recursively add 3 txns with updated amount OR P2 should insert all 3 txns and then only P1 should update the amount.
It shouldn't happen that P2 inserted 1 record with old amount and later 2 with updated amount.


I agree, chances of happening while edit is very hard, but it's possible that two people open the group at the same time and double inserts can happen.
I made lock so that this doesn't happen.

justcallmelarry:
what happens if no one opens the group for over a month? will it add two recurring entries? or will it skip one?

I generally think that there should be some sort of scheduler that takes care of adding recurring expenses rather than doing some magic when people are loading the group

neonshobhit:
> what happens if no one opens the group for over a month? will it add two recurring entries? or will it skip one?

It recursively adds all the pending txns, which means it won't skip any.


> I generally think that there should be some sort of scheduler that takes care of adding recurring expenses rather than doing some magic when people are loading the group

Agreed. I wanted to make a scheduler only but decided against it for the following reasons:
- Some groups can just a txn scheduler but never open that group. Since it's not requiring even logins, abuse can be high. So I felt if someone is opening the group, they are showing the intent that they want all the "pending expenses". Those who are not opening, can be avoided.
- This is open source + allows redistribution. So, if someone wants to set Spliit up in their own server, they will need to set up crons scheduler separately if they want to use serverless. Having server in EC2 machines would be easier though, but I didn't want to limit serverless possibility becasue of just one small feature.

neonshobhit:
Yes. But we will have to build a scheduler also. For example, we wouldn't fetch all the groups all the time. We would rather fetch only the transactions that needs to be copied (Fetch only those groups). 
But yes, internally, scheduler needs to call only this function.

neonshobhit:
@rgov Have removed archive feature from this PR and opened another PR for the same. Resolved all the comments also.

prisma/schema.prisma
```diff
@@ -53,6 +53,7 @@ model Expense {
   createdAt       DateTime          @default(now())
   documents       ExpenseDocument[]
   notes           String?
+  recurringDays  Int               @default(0)
```
rgov, line 56: Indentation/alignment does not match the lines above.

src/app/groups/[groupId]/expenses/page.tsx
```diff
@@ -104,7 +104,7 @@ export default async function GroupExpensesPage({
 async function Expenses({ groupId }: { groupId: string }) {
   const group = await cached.getGroup(groupId)
   if (!group) notFound()
-  const expenses = await getGroupExpenses(group.id)
+  const expenses = (await getGroupExpenses(group.id))
```
Comment (rgov) [inline src/app/groups/[groupId]/expenses/page.tsx:107]: Unnecessary change.

src/app/groups/recent-group-list-card.tsx
```diff
@@ -54,7 +54,6 @@ export function RecentGroupListCard({
     })
 
   const isStarred = state.starredGroups.includes(group.id)
-  const isArchived = state.archivedGroups.includes(group.id)
```
rgov, line 57: Should this change still be part of this PR?

src/components/expense-form.tsx
```diff
@@ -51,6 +51,7 @@ import { match } from 'ts-pattern'
 import { DeletePopup } from './delete-popup'
 import { extractCategoryFromTitle } from './expense-form-actions'
 import { Textarea } from './ui/textarea'
+import { AsyncButton } from './async-button'
```
rgov, line 54: This import isn't used.

src/lib/api.ts
```diff
@@ -1,7 +1,15 @@
 import { getPrisma } from '@/lib/prisma'
 import { ExpenseFormValues, GroupFormValues } from '@/lib/schemas'
-import { Expense } from '@prisma/client'
+import { Expense, RecurringTransactions } from '@prisma/client'
 import { nanoid } from 'nanoid'
+import * as z from 'zod';
+// import { setTimeout } from "timers/promises";
```
rgov, line 6: Commented out code should be removed from the PR.

src/lib/api.ts
```diff
@@ -243,7 +413,69 @@ export async function getCategories() {
 }
 
 export async function getGroupExpenses(groupId: string) {
+  const now: number = Math.floor((new Date()).getTime() / 1000)
```
rgov, line 416: This calculation `Math.floor(date.getTime() / 1000` appears in a few places, it would probably improve readability to factor this into a separate function (and the inverse `*1000` calculation) since it's not super obvious what the calculation is for, I'm guessing converting a JS timestamp from milliseconds to seconds.

rgov (changes requested):
Thanks @neonshobhit for splitting the PR. I think this is a great feature and I hope @scastiel will provide feedback now. I have some style nitpicks here.

src/lib/api.ts
```diff
@@ -27,9 +35,152 @@ export async function createGroup(groupFormValues: GroupFormValues) {
...
+      }
+    })
+    return Promise.resolve(undefined)
+  }
+  const lockId:string = randomId();
+  const receivedLockId = (await acquireLockRecurringTransaction(groupId, expenseRef || expenseId, lockId))?.[0]?.lockId
+  let recTxn;
+  if (!receivedLockId) return recTxn
+  const epochTime = Math.floor((new Date(expenseDate)).getTime() / 1000)
+  const nextAt: number = epochTime + (+expenseFormValues.recurringDays * 86400)
```
rgov, line 134: This creates the transaction exactly 24 hours after the previous transaction. It might be ever-so-slightly better to use `Math.floor(epochTime / 86400) + (1 + expenseFormValues.recurringDays) * 86400` to get midnight of the day of the next occurrence.

rgov (changes requested):

src/components/expense-form.tsx
```diff
@@ -413,6 +423,34 @@ export function ExpenseForm({
...
+                    <SelectContent>
+                      {recurringDays.map(({ key, value }) => (
+                        <SelectItem key={key} value={value}>
+                          {key}
+                        </SelectItem>
+                      ))}
+                    </SelectContent>
+                  </Select>
+                  <FormDescription>
+                    Select recursive days.
```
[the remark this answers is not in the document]
neonshobhit: We can have options like 30 days, 7 days etc. Can you please suggest other title?
or
"Repeat after:" this seems okay?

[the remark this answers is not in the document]
rgov: The Apple Calendar app uses "Repeat: <never|every day|every week|...>" which seems like a decent set of choices.

src/components/expense-form.tsx
```diff
@@ -241,6 +250,7 @@ export function ExpenseForm({
     return onSubmit(values)
   }
 
+  const recurringDays = [{ "key": "Never","value": "0"}, { "key":"weekly", "value": "7"}, {"key": "fortnightly", "value": "14"}, {"key": "monthly", "value": "30"}, {"key": "bimonthly", "value": "60"}]
```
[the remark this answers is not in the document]
neonshobhit: Yes. I'll add a fixes PR later to fix this issue.
Right now, changing to every 30 days.

src/components/expense-form.tsx
```diff
@@ -182,6 +187,7 @@ export function ExpenseForm({
           isReimbursement: expense.isReimbursement,
           documents: expense.documents,
           notes: expense.notes ?? '',
+          recurringDays: String(expense.recurringDays),
```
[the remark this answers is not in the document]
neonshobhit: I started with Number, but I thought [https://github.com/spliit-app/spliit/pull/114#discussion_r1615317502](about this as well). How to store "1st of every month"? 
Cron makes more sense. [https://crontab.guru/#0_0_1_*_*](Please check here).
Most probably I will have to change every thing from days to cron - which I can add in v2 PR.

src/lib/api.ts
neonshobhit: Sure.
Better way to do this is by using crons [as mentioned in this comment](https://github.com/spliit-app/spliit/pull/114#discussion_r1615333622). I can add this here or in V2 PR where I'll be changing everything from Days to crons?

src/lib/api.ts
neonshobhit: Yes. Done.

neonshobhit:
@rgov Resolved comments. Please check.

sir-argupta:
@neonshobhit Can you please resolve conflicts

ankit-kapur:
Would love to see this merged @neonshobhit

neonshobhit:
@sir-argupta @ankit-kapur 
Apologies, Completely missed the notifications.

Anyways it's resolved and working now.
@scastiel can you please review this PR?

src/lib/api.ts
momja: Could something like this work? [reference](https://www.prisma.io/docs/orm/prisma-client/using-raw-sql/raw-queries#queryraw). I'm not following the issue with `createIfNotExists`.

```ts
const query = await prisma.$queryRaw`
  INSERT INTO "RecurringTransactions"
    ("groupId", "expenseId", "createNextAt", "lockId")
  VALUES
    (${groupId}, ${expenseId}, 0, ${lockId})
  ON CONFLICT ("groupId", "expenseId")
  DO NOTHING;
`
```

neonshobhit: You mean to use `queryRaw` instead of `queryRawUnsafe`?

terrencetjr24:
It looked like this PR has been quite inactive and has several merge conflicts that I was having issues resolving. I took inspiration from this PR for my take on the feature^

scastiel:
Implemented in #263

PR #116, opened 2024-02-26
Minor: reorder Dockerfile layers for better cache use

jantuomi:
Move package installs and prisma client generation to before copying the `src` directory into the image. Regular source code changes in `src` do not warrant rerunning those steps.


justcallmelarry (approved):
I can attest this works as expected 👌

scastiel:
💯

Issue #98, opened 2024-02-19
Feature idea: save split mode for future expenses


jantuomi:
### Use case

I use Spliit with my SO for splitting, among others, household expenses. We have a system where one person pays a larger percentage nearly every time. The "Split mode – By percentage" is great for this. However, toggling the mode and typing in the correct percentages requires a total of 12 clicks or button presses (focusing the inputs, typing the numbers). This gets kind of annoying.

As a user I would like to save my default split mode, so that when opening the expense form, the saved split mode and numeric values would get pre-filled.

### Design

A quick design idea:
![image](https://github.com/spliit-app/spliit/assets/5064726/a15ada34-b3c6-4c8e-9e34-2b5206622c1e)

The checkbox would only appear if the mode is changed from its (current) default, in order to avoid cluttering the form.

The text could read:
- Set as default split mode (implying a cross-group setting)
- Set as default split mode **for this group** (implying that the setting only affects this group)
- Set as **my** default split mode (implying that the setting only affects me)
- Set as default **split options** (implying that in addition to the split mode, also the numeric values as stored)

or some combination/variation of these.

### Implementation

Some settings are already being stored in local storage, using a compound key of `<groupId>--<someFieldKey>`. I think this method could be used for this as well.

I can do the implementation myself 👍 Comments on the idea, especially the design, are more than welcome.

justcallmelarry:
I think it makes sense to store it in localStorage, but have no good copy suggestion to make sure that everyone know it's not for all devices etc. Maybe something like "my default split mode for this device" or something would work?

I don't think it makes sense to have someone change the default split mode and then everyones settings will change.

I am wondering what would happen in this case if someone is added to or removed from the group?

jantuomi:
Yeah, having the setting affect everyone would be pretty tricky. I think a straightforward only-me, only-this-device setting would be best. It's not a big deal to set it up on each device you use anyway.

I'll have a go at implementing this.

ChristopherJohnston:
I think this is now closed with the latest release

PR #120, opened 2024-03-01
Add "save as default splitting options" feature

jantuomi:
Related issue: #98

See issue for problem statement.

## Implementation

- Add a checkbox form field `saveDefaultSplittingOptions`
- On form submit, if `saveDefaultSplittingOptions` is set, serialize the splitting options, i.e. `paidFor` and `splitMode`, to JSON and store it with a group ID -based key in local storage. Supports all split modes. Note: `BY_AMOUNT` has special behaviour where the serialized paidFor object is set to `null`. This is because storing absolute currency values do not make much sense since expenses have different total amounts. 
- On form load, if there is no data from an existing `expense` object, try to deserialize the splitting options from local storage and use them as defaults. If that fails (missing local storage key or stale data), or `paidFor = null` as with `BY_AMOUNT`, fallback to the simple default (`splitMode = "EVENLY"`).

## Challenges

I did not completely understand the behaviour where the paidFor shares are sometimes multiplied and sometimes divided by 100, so I just tried to match that behaviour in this feature. There seems to be space for improvement there. Everything seems to work as it should though in my manual testing.

### DIFF
--- a/src/components/expense-form.tsx
+++ b/src/components/expense-form.tsx
@@ -73,7 +73,7 @@ const enforceCurrencyPattern = (value: string) =>
 
 const getDefaultSplittingOptions = (
   group: Props['group'],
-): SplittingOptions => {
+) => {
   const defaultValue = {
     splitMode: 'EVENLY' as const,
     paidFor: group.participants.map(({ id }) => ({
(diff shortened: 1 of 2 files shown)


scastiel:
@jantuomi 

> I did not completely understand the behaviour where the paidFor shares are sometimes multiplied and sometimes divided by 100

They’re not supposed to be divided by 100, only the amounts should be. But it was pretty hard to implement, and some bugs remain, which is why the percentages or shares can be divided or multiplied by 100 😅

scastiel (approved):
💯

PR #122, opened 2024-03-04
Expense list: add section for planned purchases

annalouisep:
I have a pending purchase that I made for an upcoming trip and went ahead and added it to Spliit in preparation for it hitting my account. I noticed that it shows up in the "Earlier this year" section of the expense groups and would like it marked as a future purchase until that date arrives. I've attached before and after screenshots of my proposed changes.

Before: Screenshot 2024-03-03 at 6 56 34 PM

After: Screenshot 2024-03-03 at 6 58 01 PM

### DIFF
--- a/src/app/groups/[groupId]/expenses/expense-list.tsx
+++ b/src/app/groups/[groupId]/expenses/expense-list.tsx
@@ -19,7 +19,7 @@ type Props = {
 }
 
 const EXPENSE_GROUPS = {
-  PLANNED_PURCHASES: 'Planned purchases',
+  UPCOMING: 'Upcoming',
   THIS_WEEK: 'This week',
   EARLIER_THIS_MONTH: 'Earlier this month',
   LAST_MONTH: 'Last month',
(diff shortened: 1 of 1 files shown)


justcallmelarry:
I would personally find it inconvenient if someone added like 50 future expenses and these were in the list before newly added expenses from this week. Perhaps it could be added as the first group _after_ the current week expenses rather than the first group?

Though I don't think dates in the future should go into "Earlier this year" either :)

If this get's approved my one suggestion would be to call the category "Upcoming" or something rather than "Planned purchases" as the latter carries more information, which might not be correct. "Upcoming" instead just states that the date is in the future.

scastiel:
Thanks @annalouisep! Great idea to have a section for future expenses!

I agree with @justcallmelarry, “Upcoming” might be a better label than “Planned purchases”.

annalouisep:
> call the category "Upcoming"
I like it! Updated to that verbiage.

> it could be added as the first group after the current week
I balk at the idea of breaking the ordered timeline, though you raise a good point about someone entering a ton of future entries. I left it as-is and am open to ideas!

PR #124, opened 2024-03-06
Ask for confirmation to delete an expense

### LINKED ISSUE #102: Feature request: Ask before expense deletion (see #106)
An expense can be easily deleted: Click on "Edit" followed by "Delete" and it is gone for good:

```ts
async function deleteExpenseAction() {
  'use server'
  await deleteExpense(expenseId)
  redirect(`/groups/${groupId}`)
}
```

I think that's too easy. Mistakes can (and will...) happen! We should at least pop up a dialog and ask for confirmation. This is what I expect when clicking on a red button!

I personally like the idea that groups cannot be removed from the UI. How about applying this concept to expenses, too? Rather than actually removing them from the database just flag them as deleted.

Gukkey:
Resolves #102 

image

Right now I have created a separate component for the popup which is named "delete-popup". Let me know if I have to create a more generic component first (something like "popup"). Let me also know if I have to customize the popup further in any way.


justcallmelarry:
I think this is a great addition!

I don't really have anything to say on the best practice use of components, so I'll let someone else have a say about those.

I did want to say that at least the formatting checks are going to fail, though, so you might want to run prettier before Sebastien has a look at anything (`npx prettier -w .`) in order to fix the formatting

Gukkey:
I have run my files through prettier now (thanks for your comment!), and instead of just using `<DialogClose>` I will add a `<Button>` inside of it (I think it would make the close option look like a proper button and they would have same font size) and make a PR within today. Thanks for your input.

Gukkey:
I have fixed the formatting issues and the cancel option is now a proper button now. Also, I have changed the title and description a little. 

image

src/components/delete-popup.tsx
```diff
@@ -0,0 +1,51 @@
...
+export function DeletePopup({ onDelete }: { onDelete: () => Promise<void> }) {
+  return (
+    <Dialog>
+      <DialogTrigger asChild>
+        <Button variant="destructive">
+          <Trash2 className="w-4 h-4 mr-2" />
+          Delete
+        </Button>
+      </DialogTrigger>
+      <DialogPortal>
```
shynst, line 26: You probably should remove the enclosing `<DialogPortal>` as well as the following `<DialogOverlay />`, since both elements are already included in `<DialogContent />`

So basically you are adding them twice. Just having one `<DialogOverlay />` also makes the background a little bit transparent which looks better, too.

shynst (changes requested):
Looks fine. Just one little remark, though...

src/components/delete-popup.tsx
```diff
@@ -0,0 +1,51 @@
...
+        <DialogOverlay />
+        <DialogContent>
+          <DialogTitle>Delete this expense?</DialogTitle>
+          <DialogDescription>
+            Do you really want to delete this expense? This action is
+            irreversible.
+          </DialogDescription>
+          <div className="flex gap-2">
```
shynst, line 34: Try `<DialogFooter>`instead of this `<div>`

src/components/delete-popup.tsx
```diff
@@ -0,0 +1,51 @@
...
+              type="button"
+              variant="destructive"
+              loadingContent="Deleting…"
+              action={onDelete}
+            >
+              Yes
+            </AsyncButton>
+            <DialogClose>
+              <Button variant={'secondary'}>Cancel</Button>
+            </DialogClose>
```
shynst, line 45: Looks better on on small screens, if these two are reversed - i.e. `<Button><DialogClose>Cancel</DialogClose></Button>`

src/components/delete-popup.tsx
```diff
@@ -0,0 +1,51 @@
...
+    <Dialog>
+      <DialogTrigger asChild>
+        <Button variant="destructive">
+          <Trash2 className="w-4 h-4 mr-2" />
+          Delete
+        </Button>
+      </DialogTrigger>
+      <DialogPortal>
+        <DialogOverlay />
+        <DialogContent>
```
shynst, line 28: You probably should remove the enclosing `<DialogPortal>` as well as the following `<DialogOverlay />`, since both elements are already included in `<DialogContent />`.

So basically you are adding them twice. Just having one `<DialogOverlay />` also makes the background a little bit transparent which looks better, too.

shynst (changes requested):
Looks good! Just some minor UI related stuff...

Gukkey:
Hi @shynst, thanks for your input. I have pushed a commit incorporating your suggestions.

shynst (approved):
Great! Looks fine to me.

scastiel:
Great, thanks 👏

Issue #131, opened 2024-03-27
Support: Visible transactions amounts per person


vvnwng:
The transaction only shows the full amount and doesn't show how much of each transaction is allocated to each person. 

Is there a way for me to see the amounts that the transaction has split out into? Maybe I just can't find where it is?

Thanks!

manuerwin:
Very much agree, I've been meaning to ask the same thing for awhile, for example the following is much easier to understand :)
<img width="529" alt="image" src="https://github.com/spliit-app/spliit/assets/982158/b442b8f8-b168-4811-857a-7012abe7dfc4">

ChristopherJohnston:
This could probably be incorporated into the expense summary in place of "Participant A Paid for Participant B, Participant C":

<img width="744" alt="image" src="https://github.com/spliit-app/spliit/assets/3932568/86285564-4820-4886-ace3-f64022488b54">

PR #139, opened 2024-04-11
Show the impact of an expense on the active user's balance

dcbr:
Show the impact of an expense on the active user's balance. Relates to #131.

image

Note that this branch builds upon #138 as I use that for my development environment. Once the PR is in a finished state and approved, I can rebase on main.


Carual:
This is necessary

scastiel:
Thanks @dcbr this is great! I just moved the balance on a different line (which looks better especially on mobile).

Issue #106, opened 2024-02-21
Feature request: Make transactions immutable


shynst:
Thinking more about my recent feature requests #102 and #103, the following idea could be a solution for both use cases, as well as for #15:

What about, if we would make all transactions immutable?

- if we enter an expense, it stays in the database forever
- if we delete it, the expense gets marked as deleted and will not be visible in the UI
- if we edit it, a new expense will be created
  - the old one gets marked as deleted
  - the new expense keeps a reference to the old one

I think this is easy enough to implement and we will get full transparency and a full transaction history. I can start implementation of this feature if this feature request gets approved.

For the time being this affects the backend only. How this could be reflected in the UI I'll leave open for further discussions.

Tell me, what you think abut it?

justcallmelarry:
I'm wondering if it wouldn't be better to keep a separate history table just due to the size of the transaction table already.

That also helps out for potential information that you would want stored in a log. For example `message`, if that would be something to add to explain why you are changing the expense. Or `user`, which currently is not enforced on expenses as there are not real users, and message could more or less fil the same function if people use it properly.

If we were actually working with financial transactions I think this would be very prudent to implement, but as it's just a cost-splitting app for people who trust each other, I think it would help overall performance and ease of database indexes etc to just not store more information than needed in the "main table" of the database.

For my self-hosted instance of the app the performance is likely not an issue, but might become bothersome for https://spliit.app for example

shynst:
Good suggestion @justcallmelarry!
I like it for conceptual reasons - keeping the expenses table clean and lean + adding another table with extra attributes!

However, I don't think we should be too much afraid of degraded database performance. There are other things we probably would need to take into concern as the amount of data grows, like making sure we have all indices we need, converting ids into integers, optimising our queries, and adding load balancers. Luckily it scales very well with the group id, so we can easily setup a reverse proxy and add more instances.

Issue #103, opened 2024-02-20
Feature request: transaction log (see #106)


shynst:
Writing on #102 brings to my mind another concern:

It's not that I don't trust the people I share the expenses with, but imagine a scenario where someone changes the amount of some transactions way back in the past. Not only deletion could be a problem but also changes might stay undiscovered and could be a source of discussion. Who changed what, why?

I know, having a full fledged history with rollback would be nice (see #15) but for now having a simple transaction log stored in the database would be enough.

I'm thinking of a table `TransactionLog` with 4 columns: `expenseId`, `timestamp`, `user`, `message`

Gukkey:
Hi @shynst @scastiel I would like to work on this issue but might need some suggestions. Transaction logs would be much better if they were separate for each group since having one transaction log as a whole would be messy since each group might have different people.

shynst:
Actually this is superseded by #106 which solves this issue too! So I am closing this issue.

Issue #15, opened 2024-01-02
History of a transaction


beevk:
A way to see the history of who created a transaction, updated it, what was the update, etc. 

Of course, since there's no user account, it might be a challenge to implement it right now. But maybe consider it as a feature request for the future.

LIRIKKER:
I would also like this feature +1

PR #141, opened 2024-04-14
Add basic activity log

dcbr:
This PR adds a basic implementation for an activity log in each group. Currently 4 types of activity are logged: updating the group settings and creating/updating/deleting an expense. If an active user is configured, the activity message includes the name of that active user. If #95 is approved, the "infinite scroll" feature can also be applied to this activity log (the implementation here is derived from the expense list).

image

Related issues are #15, #103 and #106.

### DIFF
--- a/prisma/migrations/20240414135355_add_activity_log/migration.sql
+++ b/prisma/migrations/20240414135355_add_activity_log/migration.sql
@@ -1,0 +1,18 @@
+-- CreateEnum
+CREATE TYPE "ActivityType" AS ENUM ('UPDATE_GROUP', 'CREATE_EXPENSE', 'UPDATE_EXPENSE', 'DELETE_EXPENSE');
+
+-- CreateTable
+CREATE TABLE "Activity" (
+    "id" TEXT NOT NULL,
+    "groupId" TEXT NOT NULL,
+    "time" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "activityType" "ActivityType" NOT NULL,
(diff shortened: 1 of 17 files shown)


ChristopherJohnston:
nice! Might also be good to have the activity history for an individual expense at the bottom of the edit expense page.

justcallmelarry:
Could be nice if the expenses were links to the expense in question!

dcbr:
@justcallmelarry They actually are linking to the expense, as long as it was not deleted in the meantime. This is denoted by the small arrows on the right (and a pointer cursor while hovering).

justcallmelarry:
@dcbr ah ok! nice! was not clear from the screenshot for me 😅 i was looking for the name of the group to be a link or similar

justcallmelarry:
This would also benefit from soft-deleting expenses, so that you can un-delete the expense from the log!

dcbr:
@justcallmelarry Yes indeed! However, I didn't want to introduce too many changes here, and just introduce a basic version that can later be extended. But I agree soft-deletion would be a nice feature for a follow-up PR :)

sh3hz:
Hi @scastiel, could you please review and merge this feature when you have a chance? This is definitely required for the app. Thanks!

ChristopherJohnston:
@scastiel I think the app will benefit greatly from this PR

shynst:
> Might also be good to have the activity history for an individual expense at the bottom of the edit expense page.
> 
Great idea!

scastiel:
Thanks a lot @dcbr, amazing contribution!

PR #143, opened 2024-04-16
Auth guard (UI)

Esinel:


ChristopherJohnston:
I'm guessing this is for a personal deployment with the PR accidentally sent here.

Nice changes to add google oidc though

quintushr:
Hi,

The app is meant to stay authentication-free to keep it simple to use, so we're not going to add auth

Thanks !

Issue #136, opened 2024-04-04
Allow amount cursor edit


tasken:
While adding or editing an expense if the amount is inputted and later the user wants to edit a digit or add a comma the app won't allow cursor input and will auto select the full number. This also happens when the user switches between apps and the input field looses focus.

![Screenshot_20240404-180924~2](https://github.com/spliit-app/spliit/assets/12686734/96746240-d3dc-4c19-bc91-e07dedfe33bd)

Under this condition, number editing, digit replace is not possible offering a bad user experience building unnecessary friction.

Ideally, users should be able to edit digits after the first input or by going back and forth different apps.

PR #144, opened 2024-04-17
change onClick to onFocus, with a slight delay for safari

justcallmelarry:
## Background
I originally wanted to make the amount field select `onFocus`, as I wanted it to be possible to choose parts of the amount.
It was then pointed out that `onFocus` does not work for Safari, since the way onMouseUp works there deselects the text again after initially selecting it.
Therefore the choice was made to go with `onClick` instead, which seemed to work out for everyone involved, and you could still move around using the arrow keys for example if you just wanted to do edits.

Along came https://github.com/spliit-app/spliit/issues/136!
So I decided to look into getting it to work better, and more in line with the original thought.

## Solution
After scouring the net I found that there is a rather simple solution for allowing this to work on Safari as well, which is to add a small delay to the selection of the input, since then the `mouseup` event will have come and gone.
I'm not sure if there are other concerns regarding the solution, but I am open to feedback on it.

### Tested on
* Safari 16.5.1
* Chrome 123.0.6312.86
* Mobile Safari iOS 17.4.1

## Issues
Closes https://github.com/spliit-app/spliit/issues/136 on merge

### DIFF
--- a/src/components/expense-form.tsx
+++ b/src/components/expense-form.tsx
@@ -327,13 +327,9 @@ export function ExpenseForm({
                           onChange(enforceCurrencyPattern(event.target.value))
                         }
                         onFocus={(e) => {
-                          {
-                            // we're adding a small delay to get around safaris issue with onMouseUp deselecting things again
-                            let target = e.currentTarget
-                            setTimeout(function () {
-                              target.select()
-                            }, 1)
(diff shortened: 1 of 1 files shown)


scastiel:
Thanks @justcallmelarry!

PR #153, opened 2024-05-12
Use placeholder for new participant input

owiber:
Instead of creating the new input with "New" as its value, set a placeholder on the input. This makes it so that when you add a participant, you still see "New" but you can immediately start typing instead of having to select/delete the initial value.

Before:
spliit before

After:
spliit after


Maxco10:
I prefer an empty value. I do not know whether in world "New" might be a real name in some languages. :)

scastiel:
Good point, way better with a placeholder, thanks for the contribution :)

PR #155, opened 2024-05-19
Add active user balance and reimbursements

araffin:
This is a proof of concept to display an overview in the expense page for the active user.
I had to adapt the `Money` component to display in read "You owes B €5" (amount is positive but need to be shown in red).

I also updated `formatCurrency` to display euro and kr as suffix (10€ vs €10).

Since this is the first time I am writing in typescript/jsx, I would appreciate any feedback =)

Screenshots:
user_balance1
user_balance2

### DIFF
--- a/src/app/groups/[groupId]/expenses/active-user-balance.tsx
+++ b/src/app/groups/[groupId]/expenses/active-user-balance.tsx
@@ -1,7 +1,8 @@
 'use client'
 import { Money } from '@/components/money'
-import { getBalances } from '@/lib/balances'
+import { Reimbursement, getBalances } from '@/lib/balances'
 import { useActiveUser } from '@/lib/hooks'
+import { Participant } from '@prisma/client'
 
 type Props = {
   groupId: string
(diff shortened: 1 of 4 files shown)


src/app/groups/[groupId]/expenses/active-user-balance.tsx
```diff
@@ -41,3 +42,82 @@ export function ActiveUserBalance({ groupId, currency, expense }: Props) {
...
+  const userReimbursements = reimbursements
+    .filter((r) => r.from === activeUserId || r.to === activeUserId)
+    .map((r) => {
+      const from = getParticipant(r.from)
+      const to = getParticipant(r.to)
+      const amount = r.amount
+      if (r.from === activeUserId) {
+        return (
+          <div key={r.from + r.to} className="text-sm">
+            You owes {to.name}{' '}
```
Comment (dcbr) [inline src/app/groups/[groupId]/expenses/active-user-balance.tsx:96]:
```suggestion
            You owe {to.name}{' '}
```

dcbr (commented):
Apart from a small typo this looks fine I think. I'm only wondering why you would add this to the "Expenses" page and not to the "Balances" page? To me it seems it would fit much better there. And if it were to be added to the "Expenses" page, I would put it at the bottom, after the expenses list. Because right now it's a bit confusing that the first thing you see on the "Expenses" page is a balance overview instead of showing the (expected) overview of expenses.

araffin:
Thanks for the review =)

>  I'm only wondering why you would add this to the "Expenses" page and not to the "Balances" page? 

That's actually the main point of this PR to have a quick overview without having to do any additional click (I'm also thinking about adding the balance overview when showing the group list).

PR #158, opened 2024-05-29
Add support for group income (= negative expenses)

shynst:
A feature which I don't need very often, but dearly miss when I need it, is to **enter a group income**.

Imagine a flight you paid for your trip with Jane and Jack gets cancelled. You are receiving $90 refund from the booking agency on behalf of the group. How would you enter that?

- What I did was to enter 2 reimbursements with $30 each: "Jane --> yourself" and "Jack --> yourself". This is not only ugly and weird (calculate splits and record cash flows in the opposite direction) but also wrong: The total group balance is not reduced by $90 as ist should be.

- Another option would be to remove the original expense completely. But this is also wrong because it messes with the transaction history.

Adding a proper income is actually very simple. It's just an **expense with a negative amount**!

In this PR I am adding it as a hidden feature, because it might confuse people or someone might not like it. To enter an income just write a negative value into the amount field. The UI clearly reflects this by replacing "expense" with "income" and "paid" with "received":

Screenshot 2024-05-29 at 15 51 15

An Income in the expenses list has a negative value and the description starts with "Received by ...":

Screenshot 2024-05-29 at 15 50 50

All splits and balance calculations remain unaffected.

### DIFF
--- a/src/app/groups/[groupId]/expenses/expense-list.tsx
+++ b/src/app/groups/[groupId]/expenses/expense-list.tsx
@@ -136,8 +136,10 @@ export function ExpenseList({
                     {expense.title}
                   </div>
                   <div className="text-xs text-muted-foreground">
-                    Paid by{' '}
-                    <strong>{getParticipant(expense.paidById)?.name}</strong>{' '}
+                    {expense.amount > 0 ? 'Paid by ' : 'Received by '}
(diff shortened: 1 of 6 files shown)


scastiel:
Smart, thanks @shynst!

PR #161, opened 2024-05-30
add activity log to expense form

ChristopherJohnston:
Added a list of activity for the current expense in the edit expense form. This makes use of the recent merge of #141, which added a basic activity log.

I have chosen to create new ExpenseActivityList and ExpenseActivityItem components rather than re-use ActivityList and ActivityItem so they can be extended in future to include, for example, a summary of what exactly was changed in the update.

This PR closes #160.

### No Activity

image

### Activity List

image

PR #162, opened 2024-05-31
Add global groups feature flag for self-hosted instances

ChristopherJohnston:
For self-hosted instances - particularly those hosted on a private server - sometimes we want a group to be visible to all users regardless of whether or not they've been given the link or have the group in their recent list.

This PR adds a GLOBAL_GROUPS_ENABLE feature flag which, when true, allows for a group to be set as global in its settings page and displays a list of global groups in the groups page.

Includes changes from my previous PR #161 

### Groups Page
image

### Group Form
image


ChristopherJohnston:
will close this and bring it back in line with main branch

PR #163, opened 2024-05-31
Add pie chart of amounts by category

ChristopherJohnston:
Initial implementation of #82. Need to work on the formatting and styling, plus addition of date ranges but it's a good start.

image


ChristopherJohnston:
To be useful as a summary, I think this also needs the ability to filter by month, between a date range, since last reimbursement, etc.

Export to pdf would also be good

Maxco10:
Good idea.

PR #165, opened 2024-05-31
add expense comments

ChristopherJohnston:
Add, view and delete comments in each expense.

This is in draft because I think more can be done on the UI. For example, only showing the delete button when hovering, formatting of the add comment form etc. Functionally I think this works fine though.

### Empty Comments List and Form

image

### Comments List (With comments)

image

### DIFF
--- a/prisma/migrations/20240531162642_add_expense_comments/migration.sql
+++ b/prisma/migrations/20240531162642_add_expense_comments/migration.sql
@@ -1,0 +1,16 @@
+-- CreateTable
+CREATE TABLE "ExpenseComment" (
+    "id" TEXT NOT NULL,
+    "comment" TEXT NOT NULL,
+    "time" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "expenseId" TEXT,
+    "participantId" TEXT NOT NULL,
+
+    CONSTRAINT "ExpenseComment_pkey" PRIMARY KEY ("id")
+);
+
+-- AddForeignKey
+ALTER TABLE "ExpenseComment" ADD CONSTRAINT "ExpenseComment_expenseId_fkey" FOREIGN KEY ("expenseId") REFERENCES "Expense"("id") ON DELETE SET NULL ON UPDATE CASCADE;
+
+-- AddForeignKey
+ALTER TABLE "ExpenseComment" ADD CONSTRAINT "ExpenseComment_participantId_fkey" FOREIGN KEY ("participantId") REFERENCES "Participant"("id") ON DELETE CASCADE ON UPDATE CASCADE;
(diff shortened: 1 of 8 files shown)


ChristopherJohnston:
added ability to edit comment (click on comment, comment from displays the comment for updating)

image

dcbr:
I really like these additions to expenses, but I'm wondering how it is displayed on the expense page (together with e.g. the changes from #161)? Is it all positioned below the expense form and isn't that too crowded? Maybe it would be better to organize all this extra information using separate tabs (e.g. an "overview" tab showing the expense form, an "activity" tab showing the activity list, and a "comments" tab showing all comments)?

ChristopherJohnston:
Currently it's set underneath the expense form (below the save, delete, cancel buttons)

I did consider tabs/pages but decided against as the expense form itself is below a set of tabs for the group and I don't see there being more than a few comments per expense so it shouldn't be too over-bearing.

If it does get messy then it should be straightforward to revisit the placement.

scastiel:
Hey @ChristopherJohnston it seems that adding a comment if there is no current participant set. I guess a workaround could be to allow anonymous comments?

Also I'm not convinced by the placement of the form, below the buttons. I don't have an easy solution in mind. I wonder if comments couldn't be in a modal or a drawer for instance...

ChristopherJohnston:
> Hey @ChristopherJohnston it seems that adding a comment fails if there is no current participant set. I guess a workaround could be to allow anonymous comments?

Quick fix is to only show the form if an active user is selected (updated and committed). With this implementation it's difficult to allow anonymous due to the foreign key constraint on participantId.

> 
> Also I'm not convinced by the placement of the form, below the buttons. I don't have an easy solution in mind. I wonder if comments couldn't be in a modal or a drawer for instance...

A modal or drawer could work. I'd have to spend a bit more time looking into how to implement this.

ChristopherJohnston:
Added Modal for comment form:

image


image


image

sir-argupta:
Hi @ChristopherJohnston 
This just my suggestion, I think we bit need to work on appearance, like buttons need to change color on hover something like those things. anyways it's nice feature to have on.

ChristopherJohnston:
@scastiel this is ready to go unless you have further comments.

ChristopherJohnston:
@scastiel bump 😀

ChristopherJohnston:
@scastiel -> solved merge conflicts from 1.14.0 update

ChristopherJohnston:
There is one thing to fix now, which I'm struggling with. The previous version used next/navigation/redirect on server side to refresh the comments after the add/edit/delete action...these now use router.refresh() which doesn't seem to have the desired effect of refreshing the page and therefore the comments.

I have tried const [comments, setComments] = useState() and then calling a refreshComments function to setComments but this seems to cause the page to crash due to too many redraws.

currently stuck..

ChristopherJohnston:
the answer was staring in my face - the tRPC useQuery hook will trigger a refresh when the data is invalidated. so adding a comments.list.invalidate({expenseId}) call in each of the add/update/delete callbacks will cause the comments list to refresh.

tRPC is nice :)

ChristopherJohnston:
@scastiel - still eager for this to be merged...

PR #168, opened 2024-06-14
Do not remove just created Docker image

### LINKED ISSUE #151: Pull access denied for spliit2 when attempting to run in a container
Was following the instructions to run in a container and running `npm run start-container` results in the following error: 
```
Error response from daemon: pull access denied for spliit2, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```
What am I missing?

iqqmuT:
`docker image prune -f` command removes just created `spliit2` Docker image. We probably want to keep the image.

Fixes #151

### DIFF
--- a/scripts/build-image.sh
+++ b/scripts/build-image.sh
@@ -8,5 +8,3 @@ docker buildx build \
     -t ${SPLIIT_APP_NAME}:${SPLIIT_VERSION} \
     -t ${SPLIIT_APP_NAME}:latest \
     .
-
-docker image prune -f


justcallmelarry:
The command `docker image prune` only removes _dangling_ images, ie images that are not tagged.
The command is there to clean up previous builds of the image, as they are no longer tagged with the `latest` tag

While I disagree that removing this will fix #151, I also think that running prune every time is good practice, as it also removes any other dangling images (not only for spliit).

TLDR: I would be completely OK with this removal, but I doubt that it will solve the issues

Issue #132, opened 2024-03-27
Floppy disk icon in Save button


iqqmuT:
The save button in group settings page has a floppy disk icon. For better user experience for all generations, some other icon could be used or it could be even without any icon.

ChristopherJohnston:
Hi iqqmuT, feel free to submit a PR with what you think is more appropriate

pktiuk:
I disagree with these statements.
Floppy disc is currently associated with saving. Young people not aware of the source of this icon often call it a save icon.  
Just write `save icon` in google.

PR #169, opened 2024-06-16
Group form: floppy disk icon removed

iqqmuT:
* floppy disk is not a modern icon
* some users might associate floppy disk icon to saving data to a local device but here we are saving data in cloud
* no icon needed, coloring indicates it is the submit button
* if save button has an icon, then cancel button should have also for symmetry

Refers #132

### DIFF
--- a/src/components/group-form.tsx
+++ b/src/components/group-form.tsx
@@ -34,7 +34,7 @@ import {
 import { getGroup } from '@/lib/api'
 import { GroupFormValues, groupFormSchema } from '@/lib/schemas'
 import { zodResolver } from '@hookform/resolvers/zod'
-import { Save, Trash2 } from 'lucide-react'
+import { Trash2 } from 'lucide-react'
 import Link from 'next/link'
 import { useEffect, useState } from 'react'
 import { useFieldArray, useForm } from 'react-hook-form'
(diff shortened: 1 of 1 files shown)


dcbr:
I don't really agree with your first two points, see for example following discussions on this topic (from 13 years ago, yet "floppy disk" save icons are still in use even today!): https://stackoverflow.com/questions/1019573/save-icon-still-a-floppy-disk and https://ux.stackexchange.com/questions/3117/save-icon-is-the-floppy-disk-icon-dead.

> * no icon needed, coloring indicates it is the submit button

Fair enough

> * if save button has an icon, then cancel button should have also for symmetry

But I like this latter approach more personally.

aloxe:
> * floppy disk is not a modern icon

That's right, it was added in Unicode for version 6.0 in 2010. Along with the 📼 Videocassette, 📤 input and output trays, 📠 fax machine, 📟 Pager and 📯 postal horn. The postal horn is maybe the oldest unused device but the symbol is still used in some countries as a sign for post offices and postal services.

iqqmuT:
Thank you for your good points. Feel free to close this pull request. Perhaps the floppy disk icon can be re-evaluated in 10 years or so. 😊

pktiuk:
I also disagree with these statements.   
Floppy disc is currently associated with saving.    
Just write `save icon` in google.

Issue #18, opened 2024-01-03
REQUEST: Add Location of Transaction


LIRIKKER:
Would be good to see the location of the transaction as you click on the expense for further details on a map. This could be by adding the location in Realtime using the devices GPS, or adding the location afterwards by providing a location.

Big-Hammer:
This could be useful for me too - I'm often using this sort of app on the road.

PR #172, opened 2024-06-19
Store locations of expenses

Teilkasko:
Inspired by #18 I implemented a first iteration for storing locations of expenses. It adds a new card to the expense form where users can view, set, change and delete the location of where an expense was made. 

So far the only way to set/change a location is by determining the current position of the user (accessing `navigator.geolocation`). Upon the first time a user uses this feature they will have to confirm the browsers dialogue asking for permission to access their position.

Locations are stored in a separate table `Point`, which has a 1:1 relationship to `Expense`, this way it is ensured that a position always consists of both latitude and longitude (or otherwise it does not exist). Initially my plan was to use the [point datatype of Postgres](https://www.postgresql.org/docs/current/datatype-geometric.html#DATATYPE-GEOMETRIC-POINTS) for the storage, but unfortunately at this point [the Prisma ORM does not support the PG geometric data types](https://github.com/prisma/prisma/issues/1798) (and apparently will not any time soon).

### Set a location

https://github.com/spliit-app/spliit/assets/31203844/0bc9a03e-46be-428e-8a9e-c10cd760c941

### Update a location

https://github.com/spliit-app/spliit/assets/31203844/abf33ef8-0c63-4b6b-8922-135149e26eab

### Remove a location

https://github.com/spliit-app/spliit/assets/31203844/39f272ba-c612-4081-8058-814027a91965

### Error while determining location 

(e. g. the user denies the location permission request)

https://github.com/spliit-app/spliit/assets/31203844/9abded1e-48d3-4252-ba03-9a762a3bd5bb


dcbr:
Awesome feature, great job!
Extra suggestion: Could you maybe add a textfield (e.g. under the expense date) that allows you to freely describe a location without having to put a marker on a map? Then you could add a button right next to that textfield that links towards the Location card to further specify it/put an exact marker.

Teilkasko:
Thank you for the feedback! @dcbr 

> Could you maybe add a textfield (e.g. under the expense date) that allows you to freely describe a location without having to put a marker on a map? Then you could add a button right next to that textfield that links towards the Location card to further specify it/put an exact marker.

Intuitively it feels a bit like creating a redundancy, as such a free-text description/mentioning of a location can already be done as part of the existing input fields (title or notes). I find myself usually writing the name of a place, where an expense was made, into the title, i. e. "Dinner at 4 seasons". Also I can imagine that for users it might be a bit confusing to have two different points in the app page where to store location-related details (the input field in the top card and then further down the location card with the map view). From your perspective, what would be the benefit to add another input field to the already existing ones?

dcbr:
My idea was that this free text location description could be used in the expense overview list (which doesn't make sense for coordinates only). I quickly drew an example on the figure below (extracted from your video).
image
That's a bit cleaner than putting the location information in the title (or notes), in my opinion. It might also make location based filtering easier if that would be added in the future.
Regarding the UI inside the expense form, we might have to fine tune that a bit. But I guess if both the text input and map view are close together, it shouldn't be too confusing.
Now that I think about it, it might be possible to extract the city name automatically from the marker location? That would also "suit my needs", although others might want to specify alternative descriptions (including street names for example, or only the country), hence why I would make it free-text. Or maybe we could make this a group setting (e.g. show country only, country + city, city only, city + street)?

Petersmit27:
Personally, I like the idea as an option but I do doubt the utility of this feature. Sure, it's cool to see exactly where the expense was made, but it takes up quite a bit of space on the expense page. I'm quite hesitant in adding more and more things to a page that should be as easy and quick to read as possible.

I think it would be good if the location block was collapsible (collapsed by default), and that it remembers the (un)collapsed state between expenses. This way, people who don't use the feature are not bothered by it and people who do use it don't have to uncollapse a block every time.

I also like @dcbr's idea to build on top of this, automatically getting a location description in the expense list. That would make me use this location feature much quicker. With a setting in the group for an appropriate format (city trip, country trip, or cross-country) this could be a really smart/frictionless feature.

But these are just my thoughts, please let me know yours :)

Issue #150, opened 2024-05-07
Feature request: Split with adjustments


rafaeldamasceno:
As also available in Splitwise, there's a mode that splits evenly with adjustments, meaning the total amount minus the adjustments you insert for each person is divided evenly between the selected person, e.g. if you make a big purchase to split evenly but someone bought something only for themselves.

hundertzwoelf:
Did anyone ever have a go at the implementation of this? I feel like this is a pretty fundamental splitting option. I was using all the time in Splitwise and found their implementation pretty intuitive to use. (I.e. after specifying the total, add how much of the total bill should only be payed by each person, then split up the remainder evenly.)

Akatrielaiic:
I would like this feature too

trincasidra:
Bumping this up. This feature alone is keeping my friend groups from moving out of splitwise, even with all the ads and bloat. I see PR #173 was merged, but it must have been since invalidated.

The way most people would envision it would probably be something like this (continuing with the original example of a 18€ expense):

<img width="1179" height="961" alt="Image" src="https://github.com/user-attachments/assets/bb61f317-a570-499c-a985-97bdbc219aaf" />

PR #173, opened 2024-07-08
feat: add auto-balancing for the amount edit

mpjk:
Somewhat related issue: #150 

## Motivation
Very recently I migrated to Spliit from Tricount, and I am missing a feature which I have found very useful when adding shared expenses. 

This occurs for example, when a statement contains a personal purchase which I only want to incur on myself, and I want the rest to be shared evenly.

## Solution

__Example__: I have a shared expense of 20€, and I want to incur a 3€ purchase only to myself and divide the rest evenly.
- In this case when the total is given, the form divides it evenly with participants by default. 
- To achieve the this, I deduct my expense of the total, divide it by the amount of participants, and then add my expense back to my own part.
    - For 4 participants, the process looks like the following:
        - $\frac{Total(€) - Personal(€)}{Participants(n)} + Personal(€)\ \rightarrow \ \frac{20-3}{4} + 3 = 7.25$
        - Then this value can be input to according row, and the rest are allocated automatically (as `4.25 €`)

To avoid errors and provide intuitive balancing on multiple participants, the autobalance occurs only on the participants, whose rows have not yet been edited.

## Issues and ideas

- This implementation does not automatically allocate a surplus/remaining cent on uneven totals, though whether this should even be automatic, is another discussion.

- My implementation reflects how I have worked with expense systems in the past, it works fine when you have the necessary amounts already calculated or arbitrarily given, but its not as intuitive as the proposal in the issue. It quickly becomes obvious that manually calculating these type of expenses is cumbersome. Despite this, I find that my solution is better than none at all
    - If wanted/necessary, I can try improve this later on, to match the behaviour of the issue.

- More so related to the existing default splitting options:
    - For example when `15` and `12` shares is selected and defaulted, this carries over to other split modes. My implementation overrides this behaviour on the amount split, but this remains for example on the percentages, where you end up with `15%` and `12%` in the mentioned situation.

### DIFF
--- a/src/components/expense-form.tsx
+++ b/src/components/expense-form.tsx
@@ -46,7 +46,7 @@ import { zodResolver } from '@hookform/resolvers/zod'
 import { Save } from 'lucide-react'
 import Link from 'next/link'
 import { useSearchParams } from 'next/navigation'
-import { useState } from 'react'
+import { useEffect, useState } from 'react'
 import { useForm } from 'react-hook-form'
 import { match } from 'ts-pattern'
 import { DeletePopup } from './delete-popup'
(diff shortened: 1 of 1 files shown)


scastiel (approved):
I was sceptical when reading the description of your PR, but when playing with it I find it very intuitive to use! :)

Thanks @mpjk!

Issue #178, opened 2024-07-19
Suggested reimbursements are not stable. They should not change after a repayment.


laszlomakk:
Hi. First of all, Thanks *a lot* for working on an open-source splitwise-alternative!

---

I've noticed that the repayment suggestions are not stable.
We had a group with the following balances:
```
my_balances1 = {
    "Alice": 307594,
    "Bob": 34005,
    "Carol": -70525,
    "Dave": 43507,
    "Erin": -68593,
    "Frank": -64524,
    "Grace": -59892,
    "Heidi": -66892,
    "Ivan": -54680,
}
```
Spliit suggested the following reimbursements:
```
len(d1)=8.
d1=[('Carol', 'Alice', 70525), ('Erin', 'Alice', 68593), ('Heidi', 'Alice', 66892), ('Frank', 'Alice', 64524),
('Grace', 'Alice', 37060), ('Grace', 'Dave', 22832), ('Ivan', 'Dave', 20675), ('Ivan', 'Bob', 34005)]
```
At this point, Grace paid Dave 22832, exactly as suggested.
I would expect this edge to simply disappear from the graph, leaving 7 reimbursement suggestions, but this is not what happened.
Instead, the graph was pretty much fully recalculated, and the reimbursement suggestions became:
```
len(d2)=8.
d2=[('Carol', 'Alice', 70525), ('Erin', 'Alice', 68593), ('Heidi', 'Alice', 66892), ('Frank', 'Alice', 64524),
('Ivan', 'Alice', 37060), ('Ivan', 'Bob', 17620), ('Grace', 'Bob', 16385), ('Grace', 'Dave', 20675)]
```
This is somewhat annoying as a user. Note specifically that the flow has been rebalanced in the graph so that Grace should now pay Dave again. (Grace already paid Dave once, so why should she pay again? Not intuitive.)

---

I see that the algorithm for [`getSuggestedReimbursements`](https://github.com/spliit-app/spliit/blob/e990e00a75fe4c9eb6955e61187fc541132732a5/src/lib/balances.ts#L85) is a bit naive at the moment.

There is a good article about more complex approaches [here](https://terbium.io/2020/09/debt-simplification/). In particular, partitioning the graph into zero-subset-sum parts ("An optimal solution") looks interesting.

However, I think there might be a smaller patch for the current approach, just to fix the stability. I will try to create a pull request for that.

PR #179, opened 2024-07-19
Make recalculation stable across repayments in suggested reimbursements

laszlomakk:
Previously, after a group participant executed a suggested reimbursement, rerunning `getSuggestedReimbursements()` could return a completely new list of suggestions.

With this change, `getSuggestedReimbursements()` should now be stable: if it returns a graph with n edges, and then a repayment is made according to one of those edges, when called again, it should now return the same graph but with that one edge removed.

The trick is that the main logic in `getSuggestedReimbursements()` does not rely on `balancesArray` being sorted based on `.total` values, only that the array gets partitioned into participants with credit first and then participants with debt last. After a repayment is made, re-sorting based on `.total` values would result in a new order hence new suggestions, but sorting based on usernames/participantIds should be unaffected.

fixes https://github.com/spliit-app/spliit/issues/178

### DIFF
--- a/src/lib/balances.ts
+++ b/src/lib/balances.ts
@@ -91,11 +91,11 @@ function compareBalancesForReimbursements(b1: any, b2: any): number {
   // positive balances come before negative balances
   if (b1.total > 0 && 0 > b2.total) {
     return -1
-  } else if (b2.total > 0 && 0 > b1.total)  {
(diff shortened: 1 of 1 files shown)


scastiel (approved):
Cool improvement. Thanks @laszlomakk!

PR #184, opened 2024-07-27
Add ability to write math expression in amount field

### LINKED ISSUE #183: Feature Request | Ability to write math expressions as Expenses.
I often want to add multiple expenses together in a single expense entry. So, I want to write math expressions (mainly addition and multiplication). Currently, writing math expressions in the amount field is not supported.

parthokunda:
When creating new expenses, currently there is no support to write mathematical expressions.
My solution includes a package  **math-expression-evaluator**. 
On change in the amount field, it tries to evaluate the string and shows **the evaluated value if correct** and **'Invalid expression' otherwise.**
The evaluated amount is shown below the input box. 
**'This is a reimbursement' is renamed to 'Reimbursement'** and shifted to the right side of the total value field.

Fixes #183


parthokunda:
I have updated the formatting issues that was failing the Prettier check.

parthokunda:
@scastiel hello, I have fixed the merge problems as well. Can you check if you are free.

antonio-ivanovski:
I played with this idea in my fork, but in practice it didn't feel as natural to use as I had hoped.

Instead, I ended up adding a simple calculator directly to the amount input, which has worked out much better while keeping the UI clean.

PR #195, opened 2024-08-03
Update splitMode selection

vosatom:
This PR updates the split mode selection to a tab-like element so users can change the mode with one click. 

If this was the reason for persisting the default mode, it is now unnecessary and therefore deleted.

screen


justcallmelarry:
Hmm, pretty sure that the storing of default splitting options also stored if you had an uneven split or similar (like 60/40 to one person due to larger income in a household group or similar)

See original PR for more details: https://github.com/spliit-app/spliit/pull/120

Probably should not just be removed.

PR #201, opened 2024-08-08
[New][Feature] !! Duplicate expense !!

Maxco10:
This pull request adds the "Clone expense" feature to Spliit projects.
The feature can be useful as shortcut for all similar transactions.

EDIT1: The cloned expense will have new Date() as date.

https://github.com/user-attachments/assets/12f30121-efe7-4b53-8b7e-0d5d93c31744


Maxco10:
Updated PR with latest change on languages

CarliJoy:
From a user perspective it is not clear what happens with the date? Why should I create a expense with the same date and time?
I would expect that the date is not copied but a new expense with the current date is created.
But than it is not really a clone rather a recreation.
Maybe in the clone dialog it would be nice to change the creation date directly with the option to set it to today?

Maxco10:
@CarliJoy 
Thank you for your feedback.
When I made the feature, I was thiking a clone of expense(all fields) but your point of view, it is correct about date.

Maxco10:
The cloned expense will have new Date() as date.

antonio-ivanovski:
For anyone waiting for this feature, i have implemented it in my fork https://github.com/antonio-ivanovski/spliit-cloud/ available at https://spliit.cloud/

Image

Image

PR #203, opened 2024-08-09
add feature flag to enable setting a group as "global"

ChristopherJohnston:
For self-hosted instances - particularly those hosted on a private server - sometimes we want a group to be visible to all users regardless of whether or not they've been given the link or have the group in their recent list.

This PR adds a GLOBAL_GROUPS_ENABLE feature flag which, when true, allows for a group to be set as global in its settings page and displays a list of global groups in the groups page.

Groups Page

image

Group Form

image

This is a re-submission of #162


ChristopherJohnston:
@scastiel thoughts on this? I think it is a useful feature for self-hosters

ChristopherJohnston:
@scastiel bump

ChristopherJohnston:
@scastiel -> solved merge conflicts from 1.14.0 update

messages/de-DE.json
```diff
@@ -56,6 +56,7 @@
     "recent": "Letzte Gruppen",
     "starred": "Favorisierte Gruppen",
     "archived": "Archivierte Gruppen",
+    "global": "Global Groups",
```
Petersmit27, line 59: In a recent update, English is used as a backoff language, so there's no need to put these labels in other languages. It also makes it clear in Weblate what has(n't) been translated yet :)

PR #205, opened 2024-08-16
i18n/add german translation

bitgroestl:


tedzards509:
I find these translation a bit more awkward than #207. However some translations are better in this one so these two translations should probably be compared to improve the final German translation.

scastiel:
Hi @bitgroestl, I merged the other PR for the German translation, but thanks for the PR. Don't hesitate to make a new one if you think translation can be improved :)

PR #206, opened 2024-08-16
fix Dockerfile

bitgroestl:
I unintentionally messed up the original pull request (#200) by deleting the my forked repo that contained the changes. That's why i'm now creating this new identical pull request. Sorry for the inconvenience, i'm still learning how Github works.

PR #208, opened 2024-08-18
Update Dockerfile

murez:
fix the image builder script error


justcallmelarry:
Also fixed in https://github.com/spliit-app/spliit/pull/206, which also fixes additional issues.

Issue #194, opened 2024-08-03
Amounts wrongly displayed


EnablSiteAdmin:
When an expense is split unevenly by specific amounts, and you revisit the expense after saving it, the original amounts you entered are not displayed. Instead, the total expense is shown divided equally among all participants. However, if you download the transactions as a JSON file, the correct original amounts are shown.

EnablSiteAdmin:
Now it's not only with split by amount, but for all advanced splitting options.

PR #210, opened 2024-08-22
Fix: Correctly display loaded expense

### LINKED ISSUE #209: Shares are not displayed correctly when editing an expense
When loading an expense (Unevenly - By shares) for editing all shares are set back to 1 when loading the form. I tried to find out what is happening here, but didn't find the cause.

When setting breakpoints in `expense-form.tsx` in line 260 and 268 I can see that the shares are correctly loaded from the database. But I get two stops on line 260 and three on line 268. With the third stop all shares are reset to 1.

I hope this description helps to find out what is happening. If needed I can provide more information.

Version: 1.9.0 (5dfe03b3f11e16e16355e65ef092da77715b4bd1)

tobiasge:
- Don't load default split options after displaying an existing expense
- Re-validate form after changing the "paidFor" selection. 
  This fixes the error message "The expense must be paid for at least one participant." after clicking "Select None" and the selecting one participant.

### DIFF
--- a/src/components/expense-form.tsx
+++ b/src/components/expense-form.tsx
@@ -256,20 +256,18 @@ export function ExpenseForm({
   }, [form.watch('splitMode'), form.watch('amount')])
 
   useEffect(() => {
-    const totalAmount = Number(form.getValues().amount) || 0
-    const paidFor = form.getValues().paidFor
     const splitMode = form.getValues().splitMode
 
-    let newPaidFor = [...paidFor]
-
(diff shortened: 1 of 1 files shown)


parthokunda:
This needs to be fixed quickly. I have already faced issues when editing my expenses, all the participants were getting added by default even though I was only trying to edit something else.

scastiel:
Thanks @tobiasge and @parthokunda!

Closes #209. Closes #194.

PR #227, opened 2024-09-25
Fix amount preview for scanned receipts

### LINKED ISSUE #87: Create from receipt preview shows wrong amount even though it is right on create page
Shows .17 should show 16.88

Probs a small display error.

<details>
  <summary>Screenshots</summary>
  
  Screenshot_20240207-200952
Screenshot_20240207-201054

  
</details> 

PS: On a side not, I feel that the receipt detector is not very good at detecting DD/MM/YYYY it usually assumes MM/DD/YYYY.
Might be nice to have either a setting for a preference or guess based on the language/location of the receipt or something like that.

### LINKED ISSUE #121: [Bug] Invoice scanning - wrong amount in preview
When scanning an invoice, an incorrect amount is displayed in the preview, although the correct amount is entered in the field.

standard-rechnung-beispiel-1024x540
Bildschirmfoto 2024-03-03 um 13 12 21
Bildschirmfoto 2024-03-03 um 13 12 31

### LINKED ISSUE #140: Amount Format Error while creating from recipe
In the public version, the Amount after scanning is a wrong formatted number.
### After scanning with GPT:
Screenshot 2024-04-11 200617 
_Example recipe: no real data_
### Real amount (after clicking continue):
Screenshot 2024-04-11 201148

The Amount gets rounded and the Cents are not right.
I cannot find out what causes this problem, maybe someone could take a look into it?

The problem could be somewhere around line 201 or in the format Function, which appears to work right in other places in the application. https://github.com/spliit-app/spliit/blob/2fd38aadd91ee5678088b00450b638086307b169/src/app/groups/%5BgroupId%5D/expenses/create-from-receipt-button.tsx#L201

mertd:
To ensure the formatter works as expected, I also set up testing. The test cases ended up taking a lot more time than the fix itself, but I wanted to make sure not to introduce any other unexpected formatting bug.

As a mostly unrelated change, I needed to change the GPT model used as I was getting depreciation errors during development.

Fixes #87 @RayBB 
Fixes #140 @Phlipde
Fixes #121 @webzit

Do the changes from #81 still work as expected @justcallmelarry?

### DIFF
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -1,0 +1,18 @@
+import type { Config } from 'jest'
+import nextJest from 'next/jest.js'
+ 
+const createJestConfig = nextJest({
+  // Provide the path to your Next.js app to load next.config.js and .env files in your test environment
+  dir: './',
+})
+ 
+// Add any custom config to be passed to Jest
+const config: Config = {
+  coverageProvider: 'v8',
+  testEnvironment: 'jsdom',
+  // Add more setup options before each test is run
+  // setupFilesAfterEnv: ['<rootDir>/jest.setup.ts'],
+}
+ 
+// createJestConfig is exported this way to ensure that next/jest can load the Next.js config which is async
+export default createJestConfig(config)
--- a/package-lock.json
+++ b/package-lock.json
@@ -56,8 +56,12 @@
         "zod": "^3.22.4"
       },
       "devDependencies": {
+        "@testing-library/dom": "^10.4.0",
+        "@testing-library/jest-dom": "^6.4.8",
+        "@testing-library/react": "^16.0.0",
         "@total-typescript/ts-reset": "^0.5.1",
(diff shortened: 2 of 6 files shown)


mertd:
@scastiel I remember why I left this branch untouched for a long time: I couldn't test if this actually fixed the issue without being able to upload to a public S3 instance. However, when I picked this up again yesterday, I assumed I had it tested manually already and just needed to finish up the automatic tests. 😅 

Could you please trigger a preview deployment with Vercel? If not, I will have to pick this up again later for some testing with a public S3 instance.

scastiel:
Hey @mertd, I just tested and it works like a charm!

I'm taking the liberty to assume your PR is "ready for review" so I can merge it (I don't have much time to dedicate to Spliit, so who knows when I have a PR-merging moment in the future 😉).

Thanks very much! (And thanks for introducing Jest too, might be useful!)

PR #231, opened 2024-09-29
Fix format for integer amounts

### LINKED ISSUE #229: Expenses are being multiplied by 100
It seems that with the new 1.10.0 release, expenses may be broken in the fact that they are all being multiplied by 100. For example if I put $0.02 and hit submit, it then shows as $2.00.

mertd:
Fixes #229


mertd:
It seems like #227 was merged a bit prematurely after all. This PR should fix the issue reported in #229. @scastiel

scastiel:
Thanks for the fix @mertd!

RayBB:
Thanks for fixing :)

PR #232, opened 2024-09-30
Fix dockerfile

Mubelotix:


totti-rdz:
FYI: duplicate of #206

Mubelotix:
Meh, someone should have merged #206 and spare me the trouble

PR #259, opened 2024-10-24
Feature: Default currency symbol

sebiweise:
In this PR I´ve added a env parameter that can be used to define a default currency symbol on group creation. The symbol can still be changed on creation by inserting another symbol into the input field.

The following variables were added:

| VAR name   |      Default      |  Description |
|----------|:-------------|------|
| NEXT_PUBLIC_DEFAULT_CURRENCY_SYMBOL |  none | This parameter can be used to define a default currency symbol that will be used on group creation. |

ToDo:
Currently the env var is defined directly via `currency: process.env.NEXT_PUBLIC_DEFAULT_CURRENCY_SYMBOL || '',` and not via the env.ts file. It is working fine, but we are missing the validation of zod. I wasn´t able to define the variable via the env.ts file, because some of the vars are missing on the clientside. Maybe we will have to seperate this config file or is there any other solution of doing this?

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,2 +1,4 @@
 POSTGRES_PRISMA_URL=postgresql://postgres:1234@localhost
 POSTGRES_URL_NON_POOLING=postgresql://postgres:1234@localhost
+
+NEXT_PUBLIC_DEFAULT_CURRENCY_SYMBOL=""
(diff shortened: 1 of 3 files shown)


sebiweise:
Maybe we could use this example to implement client, server and runtime based env validation?
https://github.com/lukevella/rallly/blob/9bbd8e0790ca0ded029ff565c9ca3cb73791fae1/apps/web/src/env.ts

scastiel:
Nice, thanks @sebiweise!

PR #260, opened 2024-10-25
[Translation] Add Traditional Chinese (zh-TW)

icarusxxy:
* Add zh-TW translation file

* Add zh-TW to other translations


scastiel:
Thanks @icarusxxy!

PR #261, opened 2024-10-29
Spliit's version in the footer

Maxco10:
Hi,

This pull request will adding a simple badge for version in the footer of project in order to display what version is using.
The version is get from `package.json`.

Each new version of the project will need to update the `package.json` with a version value.

immagine

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "spliit2",
-  "version": "0.1.0",
+  "version": "1.14.1",
   "private": true,
   "scripts": {
     "dev": "next dev",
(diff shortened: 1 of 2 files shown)


tasken:
Why not add this to the existing footer details? The way is placed now seems off the content section.

tasken:
Sorry, I'm looking at the screenshot.
I mean to align it with the existing logo and credits as an extra line. It's just a suggestion.

Petersmit27:
Hey there! This is nice, but I also think it would fit much better among the existing text in the footer. 

Do you mind if I adapt this PR and merge that? Or would you like to implement this? @Maxco10

Petersmit27:
Nice! Could you push that?

Maxco10:
> Nice! Could you push that?

Branch and pr updated!!

PR #263, opened 2024-11-06
Add recurring expense functionality

trandall2:
## What
- Closes issue #5 
  - An expense can be created with a recurrence/repeating rule, and subsequent expenses will be created
  - Deleting the most recent expense causes no further repeating expenses to be created
  - Changing the most recent expense to no longer repeat causes no further repeating expenses to be created
  - Changing the recurrence rule of the most recent expense causes all subsequent expenses to adopt the same recurrence

## Particular shortcomings to note
- Creating a monthly expense on the 29th, 30th, or 31st will not cause the expense to consistently be created for that date
  - If an expense is created on any of these 3 dates, it will eventually adopt the largest date available in the next month if the target date is not available
  - Ex 1
    - A monthly expense is initially created on January 31st, 2024
    - The next 3 expenses will be created on: February 29th, March 29th, and April 29th
  - Ex 2
    - A monthly expense is initially created on March 31st, 2024
    - The next 3 expenses will be created on: April 30th, May 30th, and June 30th
  - Ex 3
    - A monthly expense is initially created on July 31st, 2024
    - The next 3 expenses will be created on: August 31st, September 30th, and October 30th

## Acknowledged decisions
- Only supporting recurrence rules (daily, weekly, and "monthly" -- excluding the 29th, 30th, and 31st)
  - This seems like a good starting place for the feature
- Not indefinitely linking Expenses to their future related Expenses
  - It's still an option to add this linking later. But not adding it now keeps the model more flexible
- Creating repeating Expenses when a user loads all the Expenses from a group
  - I took inspiration from [this earlier PR](https://github.com/spliit-app/spliit/pull/114) which chose to do something similar. I think that it's possible/straightforward to add a job to perform this task. But this method was very lightweight and doesn't preclude a later change.

## Potential Future improvements
- Add usage of the [rrule](https://github.com/jkbrzt/rrule) library to hopefully avoid the poor end-of-month recurrence behavior mentioned above
  - I think it would be best to add this library before adding more `RecurrenceRules` (ex. bi-weekly, etc)
- Add a popup to the Expense page that notifies users about particular deletion behavior:
  - If they are deleting an Expense whose RecurringExpenseLink is pending to create the next Expense, then no further repeating Expenses will be created for that Expense
  - If they are deleting an Expense whose RecurringExpenseLink has already created the next Expense, then only the target Expense will be deleted

### DIFF
--- a/messages/es.json
+++ b/messages/es.json
@@ -136,6 +136,15 @@
         "label": "Recibido por",
         "description": "Seleccione el participante que recibió los ingresos."
       },
+      "recurrenceRule": {
+        "label": "Expense Recurrence",
+        "description": "Select how often the expense should repeat.",
+
+        "none": "None",
+        "daily": "Daily",
+        "weekly": "Weekly",
+        "monthly": "Monthly"
+      },
       "paidFor": {
         "title": "Recibido para for",
         "description": "Seleccione para quién se recibió el ingreso."
--- a/messages/es.json
+++ b/messages/es.json
@@ -137,13 +137,13 @@
         "description": "Seleccione el participante que recibió los ingresos."
       },
       "recurrenceRule": {
-        "label": "Expense Recurrence",
-        "description": "Select how often the expense should repeat.",
+        "label": "Recurrencia del gasto",
+        "description": "Seleccione con qué frecuencia debe repetirse el gasto.",
 
-        "none": "None",
-        "daily": "Daily",
-        "weekly": "Weekly",
-        "monthly": "Monthly"
+        "none": "Ninguno",
+        "daily": "Diario",
+        "weekly": "Semanal",
+        "monthly": "Mensual"
       },
       "paidFor": {
         "title": "Recibido para for",
--- a/messages/de-DE.json
+++ b/messages/de-DE.json
@@ -136,6 +136,15 @@
         "label": "Empfangen von",
         "description": "Wähle das Mitglied, das die Einnahme erhalten hat."
       },
+      "recurrenceRule": {
+        "label": "Expense Recurrence",
+        "description": "Select how often the expense should repeat.",
+
+        "none": "None",
+        "daily": "Daily",
+        "weekly": "Weekly",
+        "monthly": "Monthly"
+      },
       "paidFor": {
         "title": "Empfangen für",
         "description": "Wähle für wen die Einnahme empfangen wurde."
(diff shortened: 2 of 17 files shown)


terrencetjr24:
Test feature in action:
image

terrencetjr24:
image

messages/es.json
```diff
@@ -136,6 +136,15 @@
         "label": "Recibido por",
         "description": "Seleccione el participante que recibió los ingresos."
       },
+      "recurrenceRule": {
+        "label": "Expense Recurrence",
```
tasken, line 140: Gastos recurrentes

messages/es.json
```diff
@@ -136,6 +136,15 @@
...
+        "description": "Select how often the expense should repeat.",
+
+        "none": "None",
+        "daily": "Daily",
```
tasken, line 144: Diario

messages/es.json
```diff
@@ -136,6 +136,15 @@
...
+        "weekly": "Weekly",
```
tasken, line 145: Semanal

messages/es.json
```diff
@@ -136,6 +136,15 @@
...
+        "monthly": "Monthly"
```
tasken, line 146: Mensual

messages/es.json
```diff
@@ -136,6 +136,15 @@
         "label": "Recibido por",
         "description": "Seleccione el participante que recibió los ingresos."
       },
+      "recurrenceRule": {
+        "label": "Expense Recurrence",
+        "description": "Select how often the expense should repeat.",
+
+        "none": "None",
```
tasken, line 143: Ninguno

messages/es.json
```diff
@@ -136,6 +136,15 @@
         "label": "Recibido por",
         "description": "Seleccione el participante que recibió los ingresos."
       },
+      "recurrenceRule": {
+        "label": "Expense Recurrence",
+        "description": "Select how often the expense should repeat.",
```
tasken, line 141: Seleccione con qué frecuencia debe repetirse el gasto.

tasken (changes requested):
Adding spanish translation strings

messages/es.json
trandall2: Added translations

scastiel:
Thanks @trandall2 and sorry for the delay. It's been the most wanted feature since I launched Spliit :D

Issue #238, opened 2024-10-09
Feature Request UI: Expenses Overview: Indicate if expense has attachement


CarliJoy:
When splitting expense we always add the image for the recipes in order to have some control what was spent.

It would be helpful if there is an indicator in the UI showing if there are any attachments or not.

This would help finding expenses that still need information.

yuvarajsai:
@CarliJoy,

I don't see any "Attach file" feature already present while adding an expense.

Are you requesting for the feature where we can upload files for each expense? OR just a flag which indicates if there is an attachment corresponding to this expense?

CarliJoy:
Expenses can have attachments:

![image](https://github.com/user-attachments/assets/94693f39-f03f-44ec-8106-7fe015a50d19)

But there is no indicator inside the expenses list whenever a expense has an attachement or not:

![image](https://github.com/user-attachments/assets/2ed43933-f27f-4d7c-a58f-aedf74a69e2c)


In our shared flat we want to be transparent with our expenses so we require a photo of recipe. It would be very helpful to see in the list already which expense is missing it.

CarliJoy:
Thank for fixing @yuvarajsai

PR #267, opened 2024-12-02
feat(expense-list): Display the attachment count only when the expense includes attachments

yuvarajsai:
We can have attachments for expenses, but in expenses list we do not show if a particular expense has an attachment or not. Relates to [#238](https://github.com/spliit-app/spliit/issues/238)

_**Before:**_
image

**_After:_**

Expense has attachments:

image
image

Mobile:

image

No attachments:

image

### DIFF
--- a/messages/en-US.json
+++ b/messages/en-US.json
@@ -42,7 +42,8 @@
   "ExpenseCard": {
     "paidBy": "Paid by <strong>{paidBy}</strong> for <paidFor></paidFor>",
     "receivedBy": "Received by <strong>{paidBy}</strong> for <paidFor></paidFor>",
-    "yourBalance": "Your balance:"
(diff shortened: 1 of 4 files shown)


scastiel:
This is great @yuvarajsai, thanks a lot!

I updated two things on your PR:
1. I removed the label "attachment" as we don't have translations for all languages, and I think the icon makes it pretty explicit,
2. I optimized the query to return only the document count instead of the whole document list.

yuvarajsai:
This is my first open source PR and glad that you appreciate it @scastiel 🙌
The both changes you made make perfect sense. 

Please feel free to assign any issues in this or spliit-mobile repo. I really appreciate the idea of "spliit" being an open-source project. Thank you for this initiative and your efforts.

Issue #97, opened 2024-02-19
Feature request: show computed expense per balance


ankitbahl:
In the expense list page, there is no way to know what the computed breakdown of an expense is (Jack owes $5 to John, Alex owes $2, Bob owes 8$ to Chris). This would provide some transparency in where a balance is coming from (without having to click into each expense)

Bonus - would be nice if there was some colour indication based on who the user is (if I am John, people owing me in green, me owing shown in red, others as default)

ijm8710:
Would also love this

PR #269, opened 2024-12-04
Add computed shares per expense to fix #127

### LINKED ISSUE #127: Display uneven splitting in 'Expense' view
Fantastic project, thank you very much!

I have a (hopefully minor) feature request: it would be great if uneven shares/splits would be displayed in the expense-overview:

Reproduce:

Create an expense with uneven shares, for example:

image

No indication in the expense overview:

image

Tyde:
I feel it would be of great use to see the actual shares the currently open expense would create with the implemented shares or even with equal payment. I've included the values in brackets after the name:
grafik

To reuse the existing function to calculate the shares I've had to extract a calculateShare function in totals.ts

This would fix issue #127 and #97


scastiel:
Hi @Tyde, this is a great feature, thanks!

There seems to be a little issue in your PR, when selecting "By percentage" for an expense. The displayed amounts are not correctly multiplied by 100:

Screenshot 2024-12-07 at 11 04 46 AM

Tyde:
@scastiel you're right. I did not have enough overview over the code to see that the shares have to be multiplied by 100 for the percantage mode (and I noticed for the BY_AMOUNT mode too). I've pushed a fix now. This would just add a multiplier depending on the splitMode to the live-calculation

scastiel:
Thanks @Tyde!

PR #285, opened 2025-01-04
Fixes in polish translation

pktiuk:
Minor updates to Polish translation:
- in some contexts the phrasing was a bit misleading
- There was missing field for default reimbursement title (`ExpenseForm.reimbursement`)


scastiel:
Thanks @pktiuk!

Issue #290, opened 2025-01-05
[Bug] Validate expense amount to be != 0€


CarliJoy:
Currently you can create an expense with 0€

That is a bit annoying as it creates an expense, when pressing enter in the title accidentally.

Programs should be forgiving for mistakes and prevent them :-)

Luifr:
https://github.com/spliit-app/spliit/pull/299

yuvarajsai:
Issue as been fixed with the attached pull request and it has been merged. Please feel free to close the issue.

PR #291, opened 2025-01-06
fix the amount validation while creating an expense

yuvarajsai:
This PR is to fix this issue #290 
The current validation logic is not restricting the expense creation when the amount is "0".

After:
image

### DIFF
--- a/src/lib/schemas.ts
+++ b/src/lib/schemas.ts
@@ -52,7 +52,7 @@ export const expenseFormSchema = z
         ],
         { required_error: 'amountRequired' },
       )
-      .refine((amount) => amount != 1, 'amountNotZero')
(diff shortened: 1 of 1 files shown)


src/lib/schemas.ts
```diff
@@ -52,7 +52,7 @@ export const expenseFormSchema = z
         ],
         { required_error: 'amountRequired' },
       )
-      .refine((amount) => amount != 1, 'amountNotZero')
+      .refine((amount) => amount != 0, 'amountNotZero')
```
Luifr, line 55: Was this 1 on purpose?
Do we get something by prevent 1 cent expenses?

I assume it was a mistake, specially given the error message

scastiel:
Thanks @Luifr !

PR #299, opened 2025-01-15
validation: validate expenses to be non zero

Luifr:
Added validation to prevent zero expenses from being added


Luifr:
The package lock changes are the result of `npm audit fix`
There were 4 security issues and after the command there are zero :)

scastiel:
Thanks @Luifr, but closing as #291 had the same fix :)

PR #318, opened 2025-03-24
Fix typo in translation files

totti-rdz:
This PR corrects a misspelled property name in several language JSON files, including German. The same typo was partially addressed in PR #303 (for German only), but this PR extends the fix to all affected language files to ensure correct translations display to users.

It also fixes a minor typo in the open AI prompt in src/app/groups/[groupId]/expenses/create-from-receipt-button-actions.ts


scastiel:
Thanks @totti-rdz!

PR #320, opened 2025-03-31
docker compose: allow to build container and use interanl network

6543:
now you can do:

1. clone
2. run the app
```
cp container.env.example container.env

docker compose build
docker compose up -d
```


scastiel:
Thanks @6543!

Issue #53, opened 2024-01-19
Ability to select different currency for each transaction


vladartym:
FEATURE REQUEST: Ability to select different currency for each transaction, similar to splitwise. And choose the currency for reimbursment

scastiel:
Interesting suggestion! So that would mean:
- defining a main currency for the group (among known currencies we are able to convert)
- selecting the currency for each expense (defaulting to the group one)
- calculating the converted amount for each expense (maybe storing it in the expense as well), ideally with the rate at the date of the expense, but maybe easier with the rate at the time the expense is entered.

We’d need to call an API to get conversion rates between currencies. Any API suggestion (cheap, or even better, free) is appreciated. Note that there wouldn’t be that many API calls, if we cache the conversion rates properly.

vladartym:
> We’d need to call an API to get conversion rates between currencies. Any API suggestion (cheap, or even better, free) is appreciated. Note that there wouldn’t be that many API calls, if we cache the conversion rates properly.

Yep! Everything you mentioned is spot on! I just stumbled upon this guy: https://github.com/hakanensari/frankfurter
Gets updated everyday, I dont think we need anything more than that. https://api.frankfurter.app/latest

We can also get currency conversions based on the date of the transaction e.g. https://api.frankfurter.app/2019-01-04

mertd:
What do you think about displaying the rate in an input? Allowing to overwrite the live rate may make the app more resilient to service interruptions of the currency API (see the [recent issue at ihatemoney](https://github.com/spiral-project/ihatemoney/issues/1232)).

vladartym:
> What do you think about displaying the rate in an input? Allowing to overwrite the live rate may make the app more resilient to service interruptions of the currency API (see the [recent issue at ihatemoney](https://github.com/spiral-project/ihatemoney/issues/1232)).

Great idea!!

dmlls:
> We’d need to call an API to get conversion rates between currencies. Any API suggestion (cheap, or even better, free) is appreciated.

Might be worth taking a look at https://github.com/sal0max/currencies#features for further options and perhaps implementation details.

ashaw93:
Just to chime in, maybe instead of doing the conversion, if there are outstanding balances in other currencies, show them as another section in the balance tab. That way you can keep adding custom currencies and not have to deal with exchanges.

Big-Hammer:
[...]

PR #329, opened 2025-04-21
Add main currency code for groups

whimcomp:
This change is to be able to define an optional main currency selection to groups, for the conversion rate feature, based on [this comment](https://github.com/spliit-app/spliit/issues/53#issuecomment-1901749100).
Group form with currency field
Currency selector

Originally this PR was supposed to be part of the currency conversion feature, but the PR would have been a bit large.

### Feature

Currency code in a group influences how amounts are displayed and handled in transactions. The "amount" value is treated as the number of minor units in the currency. For example, "1500" is treated as "$15.00" if the currency is USD, but "ISK 1500" if the currency is ISK (as some currencies like ISK, JPY do not have any cents).

Choosing a currency also enables display of the amounts with the correct symbol according to the locale.
For example:

display of chinese yuan in English
display of chinese yuan in Chinese

The supported currencies are the listed currencies in Frankfurter as of this 20th April 2025.

The "currency-list" package is used to generate the json containing all the currency data as well as all the translations.
The json data can be regenerated with npm run generate-currency-data (useful when another currency or language is added in the future)

### Groups without a currency code

The currency code will be undefined for all previous groups. If it is undefined, the currency symbol is required and will be used for display.
All displays will be done with a "fictional" currency defined with 100 minor units (cents) for a major unit (like Euros), this was the previous behaviour.

If the original amounts in the group were in a currency with cents, adding a currency code with cents should not have any problems.
However, if the original amounts were in a currency without cents, adding a currency will result in the amounts being 100 times higher.
For the same reason, if a currency without cents was used, and the currency code is then set to Custom, the amounts will be 100 times lower.

I haven't found a satisfying solution other to recommend not to change between currencies with different lengths.

Let me know what you think!

### DIFF
--- a/messages/en-US.json
+++ b/messages/en-US.json
@@ -94,6 +94,11 @@
       "placeholder": "$, €, £…",
       "description": "We’ll use it to display amounts."
     },
+    "CurrencyCodeField": {
+      "label": "Main currency",
+      "description": "All amounts and balances will be in this currency. Changing this will NOT convert expenses already entered.",
+      "customOption": "Custom"
+    },
     "Participants": {
       "title": "Participants",
       "description": "Enter the name for each participant.",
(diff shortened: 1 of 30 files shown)


caramelmelmel:
let the group add different currencies
people travelling with folks from different countries also matters here

PR #330, opened 2025-04-21
Specify currency per transaction

whimcomp:
This implements currency and conversion rate per expense (issue: #53), building on the previous PR #329 adding group main currency.

Expense page with currency

The rate is obtained from Frankfurter, client-side, for now (so no cache aside from what swr provides yet).
The rate can also be manually provided per expense. It is saved as a Decimal.

For convenience, if the expense is split by amount, additional inputs can be used to convert these split amounts on the fly. These values are not saved.

Convert split amounts

If the group has a custom currency (default for all previous groups), the feature won't be available.

Expense page with custom currency


antonio-ivanovski:
Does it make sense to make the groups support multi currency, so instead of converting at time of creating the expense, the group keeps each of the different currency expenses and at the end of the day the group can decide to settle based on the original currency or by converting the currencies at the time of the settling.

I agree that is also useful to have support for immediate currency convert at the time when it is added.

OnyXxL:
I agree with @antonio-ivanovski.
Let's say a traveler pays sometimes in EUR by credit card attached to an European account, and sometimes in USD from an US account.
The user at the origin of the expense could optionally tick a "Convert" box.
If the box is unchecked, the expense stays in the alternative currency selected by the user without conversion. In that case there will be a general expense balance for each currency.
If the box is ticked, the expense is converted into the main currency as it is actually.
I can't find that feature in any of the split apps on the market.

Petersmit27:
Thanks for this feature @whimcomp! :)

Petersmit27:
For the issue making the conversion optional, see #412.

I also made an issue to address a small problem I had with the split unevenly option (#411)

PR #341, opened 2025-05-18
Add Japanese translation

sora-grayscale:
Hi,
This commit adds a Japanese translation for the project content


Ancurio:
Small nitpick: I believe the language code for Japanese is `ja-JP`; it's one of those rare cases where the language code (`ja`) is not identical to the country code (`jp`).

PR #377, opened 2025-07-02
fix: Resolve 500 error on CSV export for group names with umlauts

Uli-Z:
Previously, exporting groups with German umlauts in their names resulted in a 500 error due to incorrect handling of the  header. This commit corrects the usage of the  library to properly generate both ASCII () and UTF-8 () compatible filenames, ensuring successful CSV exports for all group names.


quintushr:
Thanks @Uli-Z  for the fix!

I tested it locally with a group name containing a special character (`TestÄ`) and it works as expected.

**Proof of test** 

Screenshot from 2026-08-10 22-25-16


— the exported file is correctly named:
`Spliit Export - TestÄ - 2026-08-11.csv`

I also confirmed that the same export is still broken on the current production app, so this definitely fixes it.

Merging now — thanks again for the contribution!

PR #385, opened 2025-07-23
Localization improvements

scollovati:
1. add new message notInvolved in ExpenseCard
2. remove expense from the Recurrence field in order to make it suitable also for an income
3. add italian translation for recurrentRule
4. fixed recurrentRule json position


benjuade:
For the italian translation of recurrence is related to https://github.com/spliit-app/spliit/pull/328, while for the german tranlsation is related to https://github.com/spliit-app/spliit/pull/382 (that introduces also a new localized string, different from the one implemented here).

All useful features that need to be merged.

PR #389, opened 2025-08-02
chore(deps): bump prisma to 6.13.0

qvalentin:


next.config.mjs
```diff
@@ -25,14 +25,17 @@ if (process.env.S3_UPLOAD_ENDPOINT) {
...
   experimental: {
     serverActions: {
-        allowedOrigins: ['localhost:3000'],
+      allowedOrigins: ['localhost:3000'],
     },
-},
+  },
+  typescript: {
+    ignoreBuildErrors: true,
+  },
```
Petersmit27, line 38: Why did you add this? I don't think we'd want to ignore this

Peronia: Hi, I'm not @qvalentin or have technical insights in this pr, but I look forward to use spliit with nix.
he wrote something [here](https://github.com/NixOS/nixpkgs/pull/394621#issuecomment-2906814991) and [here](https://github.com/NixOS/nixpkgs/pull/394621#issuecomment-2994285239).

Issue #401, opened 2025-09-01
Ordering of group participants


Petersmit27:
### Problem
Group participants are ordered in the way they are created, and can never be rearranged. This makes group creation a hassle, or leaves the user with a list of participants in a random order.

### Easy solution
Participants should always be ordered alphabetically. During creation, the user can enter participants in whichever order they want, and after creation the participants are automagically sorted.

### More difficult solution
Sort participants alphabetically by default, but give the user the choice to rearrange the participants when creating/editing the group.

PR #416, opened 2025-09-14
Added Support for Reordering and Sorting Participants

archie46:
PR For Feature Request #401


archie46:
Hi @Petersmit27 ,
Sorry for the delay in getting back to you on this — I’ve now addressed the changes you suggested earlier. I’ve opened the updated PR and would really appreciate it if you could take another look when you have time. Thanks a lot for your patience!

PR #417, opened 2025-09-14
Fix decimal error in expense form

### LINKED ISSUE #414: Cannot split unevenly by amount in some conditions
Image

Image

I think this is due to the fact that the expense/(n. of participants) is not an exact number.

Petersmit27:
Closes #414 

This fix happens because of a decimal error when adding up each `paidFor` amount. In Javascript, 2.3 + 2.3 + 2.8 equals 7.3999999999999995 

This is quite a hacky way of fixing it so any feedback would be appreciated.


Petersmit27:
The problem is that we need the group's currency, not the expense's currency, because the "uneven - by amounts" option calculates (and stores) in the converted currency. 

Maybe we could pass the group's currency as a `groupCurrency` field in the expense form's model (not sure what the precise terminology is). That does feel a bit hacky though, since the group's currency shouldn't be editable or sent with the request.

Btw, no worries about introducing the bug; I just didn't test it very well.

whimcomp:
Ah, right. If the issue is (as I understand it) summing up the share amounts which are Javascript `number` doubles and falling into floating-point arithmetic traps, we could use the `Decimal` class used by Prisma in the `paidFor.shares` schema, and use its methods (`.add`, `.equals` etc.) which will perform the exact calculations and comparisons. That way we wouldn't need the group currency for this check at least.

I'm saying "at least" because there might be cases where the group currency would be needed for other potential future checks? Like checking whether the user-input amounts don't have too many decimal digits.

Issue #186, opened 2024-07-27
User data deletion


54m4d:
Hi! I really like Spliit! However, How can the user delete a group and/or its data after an event has finished?

This is important; otherwise, the group/data will still be there if the user visits the shared URL again.

There is #9. However, there is not much information on how groups/data are deleted.

To be honest, this is preventing me from continuing to use Spliit and I hope it can be sorted out soon.

Thanks!

lorenzoalali:
I agree completely. I was very surprised not to find a way to delete a group once I was finished with it. I had to manually delete all the entries and anonymize all the information.

If the concern is that anyone could delete the group because there's no authentication, that's not a valid objection in my opinion, because anyone with malicious intent who gains access to the group can already delete everything.

What's even more troubling is that https://spliit.app/groups/xxx/activity keeps the names of deleted entries. This is by design, but until we have a proper login system (which requires more effort to implement than a simple delete button, and I am not expecting this to be on the roadmap), we urgently need a way to delete a group, as keeping entered data without a way to remove it is against good privacy practices, GDPR and other regulations.

In the meantime, thanks to the developers for making this tool!

Monoxide8969:
Yeah, definitely a needed feature. Would be good for privacy and also reduce hosting costs I assume

PR #420, opened 2025-09-17
feat: schedule group for deletion

HerrLevin:
Since #9 has been open for almost 2 years and #186 requests similar functionality, I decided to build a delete function.

With this PR, users are able to schedule a group for deletion, giving them a 30 day grace period to cancel the deletion. This way, users with malicious intent will not be able to instantly delete groups.
There's also a small safety mechanism, forcing users to input the group name. That way, there's no accidental deletion of the wrong group and brute forcing is significantly harder.

One thing that still needs work is the deletion of already uploaded images into the s3 bucket. Help with this is highly appreciated.

(PS: Thank you so much for this app!)

### DIFF
--- a/messages/de-DE.json
+++ b/messages/de-DE.json
@@ -119,7 +119,15 @@
       "saving": "Speichert…",
       "create": "Erstellen",
       "creating": "Erstellt…",
-      "cancel": "Abbrechen"
+      "cancel": "Abbrechen",
+      "Delete": {
+        "title": "Gruppe löschen?",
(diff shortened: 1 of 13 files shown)


Petersmit27 (changes requested):
Nice idea and implementation :+1: I do have some comments

XSpielinbox:
For sure this implementation may not be perfect yet, but I would highly appreciate it, if some group deletion support would be merged soon. One can improve this afterwards as well in my opinion.
Having the ability to delete the data one has input would be one of the most important features in my opinion to make this a viable software. Thanks so far for the work!

PR #425, opened 2025-09-20
bugfix: Store originalAmount in minor units to match amount and avoid issues with fractional values in int database fields.

yllar:
`originalAmount` was previously not stored in minor units, causing data loss when saving fractional values since the database field is `int`. This change aligns `originalAmount` with `amount`, ensuring consistent storage and behaviour.


antonio-ivanovski:
This bug is already fixed in my fork, so edited expenses correctly retain and display their original currency.

As a bonus, the Spliit import feature also works around this issue for existing groups by reconstructing the original amount from the stored converted amount and exchange rate, preserving the original currency as expected. Feel free to give it a try!

Issue #424, opened 2025-09-20
Wrong share(s) after saving when "Unevenly - By shares" is used


wwithoutn:
When creating an expense using the "Unevenly - By shares" method, if one user's shares are customized, all other users who were left at the default value (1 share) have their shares incorrectly saved as 0.01.

When creating an expense using the "Unevenly - By shares" method, if one user's shares are customized, all other users who were left at the default value (1 share) have their shares incorrectly saved as 0.01.

<img width="765" height="616" alt="Image" src="https://github.com/user-attachments/assets/3e1ba905-0940-4e7e-b48e-1ccc34b9eebd" />

<img width="752" height="615" alt="Image" src="https://github.com/user-attachments/assets/6a22ac5b-acb2-45a8-b231-7c7a995d2aa9" />

# Replication

1. Create a new expense (e.g., $500). Choose to split mode to `Unevenly - By shares`
2. For one person, change their portion to `2 shares`. Leave the other participants at the default value of `1 share`.
3. Save the expense.
4. Re-open the newly created expense.

## Expected Result:
The participants whose shares were not changed should still be recorded as having 1 share.

## Actual Result:
The participants who were left at the default of 1 share are now recorded as having 0.01 shares.

PR #429, opened 2025-09-27
bugfix: update expense-form.tsx to handle shares as strings

yllar:
Potential fix for #424

The issue is in the data flow between the form and the schema transform function:

1. **When editing existing expenses:** Form loads shares by dividing database values by 100 (e.g., 200 / 100 = 2), but loads them as **numbers**
1. **When users change values:** Input fields return **strings** via `enforceCurrencyPattern`
1. **Schema transform:** Only multiplies by 100 for **string** values, not number values
1. **Result:** Modified shares (strings) get multiplied by 100, unmodified shares (numbers) stay as-is

**Proposed fix:** handle all shares consistently as **strings** throughout the form.

Did some limited testing and it seems to work in my dev instance, but someone with deeper knowledge on the project should review

### DIFF
--- a/src/app/groups/[groupId]/expenses/expense-form.tsx
+++ b/src/app/groups/[groupId]/expenses/expense-form.tsx
@@ -81,7 +81,7 @@ const getDefaultSplittingOptions = (
     splitMode: 'EVENLY' as const,
     paidFor: group.participants.map(({ id }) => ({
       participant: id,
-      shares: 1,
+      shares: "1", // Use string to ensure consistent schema handling
     })),
   }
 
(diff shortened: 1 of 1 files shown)


derekl-beep:
hey just a heads up - the fix works perfectly in dev, but the deployment fails to compile during `npm run build` with the following type error:

```
Type error: Type '{ participant: string; shares: string; }[]' is not assignable to type '{ shares: number; participant: string; originalAmount?: string | undefined; }[]'.
  Type '{ participant: string; shares: string; }' is not assignable to type '{ shares: number; participant: string; originalAmount?: string | undefined; }'.
    Types of property 'shares' are incompatible.
      Type 'string' is not assignable to type 'number'.
```

Petersmit27:
This is merged by merging #453. Closing this one.

PR #453, opened 2025-11-05
bugfix: Fix share values being incorrectly divided by 100 in expense form

derekl-beep:
This PR builds on #429 by adding type assertions to address TypeScript errors in the expense form. It potentially resolves issue #424.

------

## Problem

When creating or editing expenses with split modes `BY_SHARES`, or `BY_PERCENTAGE`, share values were sometimes incorrectly divided by 100 upon saving. For example:
- A user enters a share value of `1`
- After saving and reopening, it displays as `0.01`
- Subsequent saves compound the error: `0.01` → `0.0001`

This bug occurred specifically when:
1. Editing an existing expense without modifying the share values
2. Loading and using default splitting options from localStorage
3. Creating a new expense with default share values (not manually edited)

## Root Cause

The issue was caused by **inconsistent type handling** in the form and schema transformation:

- **String values** (user input) → multiplied by 100 → correctly saved as basis points ✅
- **Number values** (loaded from DB/localStorage) → bypassed transformation → incorrectly saved as-is ❌

The schema transformation used a type-based check that only applied the `* 100` multiplication to string values, assuming numbers were already in storage format. However, form initialization converted storage format (basis points) to display format (decimals) as numbers, which then bypassed the transformation on save.

## Solution

Ensure **all share values in the form are consistently strings** in display format:

1. **Convert DB values to strings** when loading existing expenses
2. **Convert localStorage values to strings** when loading default splitting options  
3. **Assign default values as strings** when initializing new forms or adding participants
4. **Keep schema transformation type-based** to only transform string values (user input)

This guarantees that all share values follow the same flow:
- Display format string (e.g., `"1"`) → Schema transformation → Storage format number (e.g., `100`)

### `src/app/groups/[groupId]/expenses/expense-form.tsx`
- Convert DB-loaded share values to strings: `(shares / 100).toString()`
- Convert localStorage-loaded share values to strings: `(paidFor.shares / 100).toString()`
- Changed default share assignments from `1` to `'1'` in:
  - Initial default splitting options
  - Reimbursement defaults
  - "Select All/None" button functionality
  - Individual participant checkbox selection

## Testing

- [x] Build succeeds with no TypeScript errors
- [x] Manually tested scenarios:
  - Create expense with BY_SHARES, save without editing → values preserved ✅
  - Edit existing expense without modifying shares → values preserved ✅
  - Edit existing expense by modifying shares → values preserved ✅

### DIFF
--- a/src/app/groups/[groupId]/expenses/expense-form.tsx
+++ b/src/app/groups/[groupId]/expenses/expense-form.tsx
@@ -81,7 +81,7 @@ const getDefaultSplittingOptions = (
     splitMode: 'EVENLY' as const,
     paidFor: group.participants.map(({ id }) => ({
       participant: id,
-      shares: "1", // Use string to ensure consistent schema handling
+      shares: '1' as any, // Use string to ensure consistent schema handling
     })),
   }
 
@@ -113,7 +113,7 @@ const getDefaultSplittingOptions = (
     splitMode: parsedDefaultSplitMode.splitMode,
     paidFor: parsedDefaultSplitMode.paidFor.map((paidFor) => ({
       participant: paidFor.participant,
-      shares: (paidFor.shares / 100).toString(), // Convert to string for consistent schema handling
+      shares: (paidFor.shares / 100).toString() as any, // Convert to string for consistent schema handling
     })),
   }
 }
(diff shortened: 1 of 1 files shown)


Petersmit27:
Thanks for the fix @yllar and @derekl-beep! 

I'm not really a fan of all the `as any`s, but I guess that this is a problem with the schema and the fact that the same schema is used in both the frontend and backend. 

I'd love to have this restructured and have a test suite to make sure these things don't break in the future...

PR #460, opened 2025-11-11
fix(db): Correct local db script volume mount for modern postgres images

### LINKED ISSUE #459: [Bug] issue with `start-local-db.sh` script and newer Docker versions
### Bug Description

I've observed a potential issue when using the `scripts/start-local-db.sh` script with what appears to be newer versions of the official PostgreSQL Docker image (e.g., version 18+). This can make testing the local development setup a bit annoying, as the database container may not start as expected.

The issue seems to be related to the volume mount point specified in the `docker run` command within the script, which might not align with the expectations of the newer postgres images.

### Steps to Reproduce

1.  Using a recent version of Docker (I observed this with version 18 of the postgres image).
2.  Run the `scripts/start-local-db.sh` script.
3.  Observe the Docker container logs for the `spliit-db` container.

### Expected Behavior

The PostgreSQL container should start successfully and be ready to accept connections for local development.

### Actual Behavior

In my observation, the `spliit-db` container started and then immediately exited with an error.

### Error Message

The container logs showed the following error, which suggests a problem with the data directory mount:

```
Error: in 18+, these Docker images are configured to store database data in a

       format which is compatible with "pg_ctlcluster" (specifically, using

       major-version-specific directory names).  This better reflects how

       PostgreSQL itself works, and how upgrades are to be performed.


       See also https://github.com/docker-library/postgres/pull/1259⁠


       Counter to that, there appears to be PostgreSQL data in:

         /var/lib/postgresql/data (unused mount/volume)


       This is usually the result of upgrading the Docker image without

       upgrading the underlying database using "pg_upgrade" (which requires both

       versions).


       The suggested container configuration for 18+ is to place a single mount

       at /var/lib/postgresql which will then place PostgreSQL data in a

       subdirectory, allowing usage of "pg_upgrade --link" without mount point

       boundary issues.


       See https://github.com/docker-library/postgres/issues/37⁠ for a (long)

       discussion around this process, and suggestions for how to do so.

```

Uli-Z:
### Description

This PR provides a minimal fix for the `scripts/start-local-db.sh` script, which was failing when used with modern PostgreSQL Docker images (version 18+).

The root cause of the failure was an incorrect volume mount point in the `docker run` command. The script was attempting to mount the PostgreSQL data volume to `/var/lib/postgresql/data`, while newer PostgreSQL images expect the mount point to be at `/var/lib/postgresql`. This discrepancy caused the container to exit with an error, preventing local database setup.

This commit addresses the issue by:
- Changing the volume mount in `scripts/start-local-db.sh` from `/var/lib/postgresql/data` to `/var/lib/postgresql`.

This is a targeted and minimal fix that resolves the immediate problem of the script failing. Other aspects of the script, such as its container management logic, remain unchanged to keep this PR focused.

### Related Issue

Fixes #459

### How to Test

1.  Ensure you have a recent version of Docker installed.
2.  Run the `scripts/start-local-db.sh` script.
3.  Verify that the `spliit-db` PostgreSQL container starts successfully and remains running.
4.  You can check the container's status with `docker ps` and its logs with `docker logs spliit-db`.

### DIFF
--- a/scripts/start-local-db.sh
+++ b/scripts/start-local-db.sh
@@ -6,6 +6,6 @@ else
     echo "postgres is not running, starting it"
     docker rm postgres --force
     mkdir -p postgres-data
-    docker run --name spliit-db -d -p 5432:5432 -e POSTGRES_PASSWORD=1234 -v "/$(pwd)/postgres-data:/var/lib/postgresql/data" postgres
+    docker run --name spliit-db -d -p 5432:5432 -e POSTGRES_PASSWORD=1234 -v "/$(pwd)/postgres-data:/var/lib/postgresql" postgres
     sleep 5 # Wait for postgres to start
 fi


1mS3r (approved):
I came to see if this was solved as this was the fix I had to apply in local, so ✅

PR #462, opened 2025-11-14
Fix/refactor expense calculation

### LINKED ISSUE #374: Rounding drops a cent when splitting evenly
When you split an amount that can’t be divided cleanly to two-decimal precision (e.g. $1 ÷ 3), each share is rounded down ($0.33), so the shares add up to less than the original amount ($0.99). That missing cent isn’t assigned to anyone, leaving the totals visibly off. While $0.01 won't ruin anyone, it still feels a bit weird. ;) 

Image

Proposal:  Hand those leftover cents out deterministically but variably so one person isn’t always stuck: simplest rule → give the remainder to the payer of that specific expense.

Uli-Z:
## Summary
- Fix #374: Every expense now distributes its full amount, even when minor units do not divide cleanly. Shares are computed with `decimal.js`, truncated toward zero to whole cents, and any remainder is distributed afterward.
- Remainder distribution by fractional parts: leftover cents (positive/negative) are allocated one by one to participants with the largest truncated fractional parts; on ties, later participants (by original order) receive the cent first.
- BY_AMOUNT stays deterministic: if submitted amounts do not sum exactly to the total, the payer gets the difference credited/debited.
- Balances and CSV export use the centralized `calculateShares` logic, keeping UI, exports, and summaries consistent.
- Comprehensive tests cover the new logic and the tie-break rules.
- The “Your share” preview in the form uses the new calculation logic.

### Refactoring Details

This PR goes beyond a simple bug fix and constitutes a significant refactoring of a core piece of the application's business logic.

**Before:**

*   **Fragmented Logic:** Expense share calculations were implemented independently and inconsistently in multiple places:
    *   The balance calculation in `src/lib/balances.ts`.
    *   The CSV export generation in `src/app/groups/[groupId]/expenses/export/csv/route.ts`.
    *   A simplistic `calculateShare` helper in `src/lib/totals.ts`.
*   **Inconsistent Behavior:** Each implementation had its own way of handling rounding and division, leading to potential bugs and inconsistencies between what users saw in the balance sheet versus what they got in an export.
*   **Precision Issues:** Calculations were performed using standard JavaScript numbers, making them susceptible to floating-point precision errors.

**After:**

*   **Single Source of Truth:** All calculation logic is consolidated in `calculateShares` (src/lib/totals.ts).
*   **Guaranteed Precision:** All internal arithmetic uses `decimal.js` for precise, finance-safe calculations.
*   **Centralized & Consistent:** `balances.ts`, `export/csv/route.ts`, and the form preview all use the central function — the same numbers everywhere.
*   **Deterministic & Testable:** Clear deterministic remainder distribution with a tie-break rule; a comprehensive test suite (`totals.test.ts`) locks behavior.

## Testing
- `npx jest src/lib/totals.test.ts --runInBand`
- Includes tie-break tests for equal fractions in EVENLY, BY_SHARES, and BY_PERCENTAGE (positive and negative amounts).

## Explanation of Remainder Distribution
- Compute each participant’s exact share using the selected split mode (evenly, by shares, by percentage, by amount).
- Truncate each exact share toward zero to whole cents. This guarantees the rounded shares never exceed the total amount.
- Compute the remainder: `remainder = totalAmount - sum(truncatedShares)`.
- Special case BY_AMOUNT: if the submitted amounts don’t sum to the total, give the entire remainder to the payer (credit or debit).
- Otherwise, distribute the remainder one cent at a time:
  - For each participant, compute the fractional part magnitude from their exact share (the portion that was cut off by truncation).
  - Order participants by larger fractional parts first. If multiple participants have the same fractional part, later participants (by the original list order) come first.
  - Add +1 cent to the first participant in this order for a positive remainder (or -1 cent for a negative remainder), move to the next participant, and repeat until the remainder is zero. 

#### Split $1.00 evenly among 3 people (P1, P2, P3)
  - Exact shares: $1.00 ÷ 3 = $0.333333… each.
  - Truncate to cents (toward zero): $0.33, $0.33, $0.33.
  - Remainder: $1.00 − ($0.33 + $0.33 + $0.33) = $0.01.
  - Tie-break on equal fractions: later participants get leftover first.
  - Distribute remainder (one cent):
      - +$0.01 → P3
  - Final amounts: P1 = $0.33, P2 = $0.33, P3 = $0.34 (sum = $1.00).

#### Split $1.00 between 2 people with shares 2:1 (P1:P2)
  - Exact shares: $1.00 × (2/3) = $0.666666… for P1, and $1.00 × (1/3) = $0.333333… for P2.
  - Truncate to cents (toward zero): P1 = $0.66, P2 = $0.33.
  - Remainder: $1.00 − ($0.66 + $0.33) = $0.01.
  - Who gets the extra cent? The one with the larger fractional part (P1: .0066… vs P2: .0033…).
  - Final amounts: P1 = $0.67, P2 = $0.33 (sum = $1.00).

### DIFF
--- a/src/app/groups/[groupId]/expenses/expense-form.tsx
+++ b/src/app/groups/[groupId]/expenses/expense-form.tsx
@@ -915,6 +915,8 @@ export function ExpenseForm({
                                             Number(form.watch('amount')),
                                             groupCurrency,
                                           ), // Convert to cents
+                                          expenseDate:
+                                            form.watch('expenseDate') ?? new Date(),
                                           paidFor: field.value.map(
                                             ({ participant, shares }) => ({
                                               participant: {
@@ -940,8 +942,15 @@ export function ExpenseForm({
                                           splitMode: form.watch('splitMode'),
                                           isReimbursement:
                                             form.watch('isReimbursement'),
+                                          paidBy: {
+                                            id:
+                                              form.watch('paidBy') ??
+                                              field.value[0]?.participant,
+                                            // name is required for the helper but irrelevant here
+                                            name: '',
+                                          },
                                         }),
-                                        locale,
+                                        locale
                                       )}
                                       )
                                     </span>
--- a/src/app/groups/[groupId]/expenses/expense-form.tsx
+++ b/src/app/groups/[groupId]/expenses/expense-form.tsx
@@ -916,7 +916,8 @@ export function ExpenseForm({
                                             groupCurrency,
                                           ), // Convert to cents
                                           expenseDate:
-                                            form.watch('expenseDate') ?? new Date(),
+                                            form.watch('expenseDate') ??
+                                            new Date(),
                                           paidFor: field.value.map(
                                             ({ participant, shares }) => ({
                                               participant: {
@@ -950,7 +951,7 @@ export function ExpenseForm({
                                             name: '',
                                           },
                                         }),
-                                        locale
+                                        locale,
                                       )}
                                       )
                                     </span>
--- a/src/app/groups/[groupId]/expenses/export/csv/route.ts
+++ b/src/app/groups/[groupId]/expenses/export/csv/route.ts
@@ -1,4 +1,5 @@
 import { getCurrency } from '@/lib/currency'
+import { calculateShares } from '@/lib/totals'
 import { formatAmountAsDecimal, getCurrencyFromGroup } from '@/lib/utils'
 import { Parser } from '@json2csv/plainjs'
 import { PrismaClient } from '@prisma/client'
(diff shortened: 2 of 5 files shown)


antonio-ivanovski:
Thanks @Uli-Z for the breakdown of the issue, i was dealing with this issue in my fork of this project at https://github.com/antonio-ivanovski/spliit-cloud
Already have merged the fix, give it a try.
One of the issues that i noticed in your proposal is that it computes the cents per expense and then uses these pre-computed in the total balances. This leads to the incorrect cents leaking into the total balance calculation. More correct approach is to make the total calculation on all the cents and the all shares. This way if there is two expenses that lead to even split, the final split is even without having to deal with "who gets the excess cent".

Another improvement is the way the cents are being distributed. It is based on round-robin but the seed is based on the stable expense id. This makes the per-expense cent distributiion more fair.

Give it a try at https://spliit.cloud/ and let me know how the calculations look to your sharp eye.

PR #468, opened 2025-11-15
Implementação de CI/CD com SonarCloud

joao-stempniak:
Create pull request


BastiOfBerlin:
Hi,

first of all, comments should be in English. And it looks to me that the project key is bound to your personal account. In order to discuss Sonar as a build action, I'd like you to supply some content to the PR description so that we can begin a discussion on the topic.

Thanks for getting involved!

PR #479, opened 2025-12-06
Upgrade dependencies

scastiel:
Migrate to latest versions of Next.js, React, Radix, etc.


glacasa:
Hi, thanks for the update. 
Is this a fix for the critical React vulnerability from last week ?

I'd like to update my instance (using Docker), but it seems the Docker image hasn't been updated yet.

PR #494, opened 2026-01-25
feat: add QR code support for sharing and joining groups

theo-engels:
Add QR code functionality to improve group sharing and joining experience:

deployment URL: [https://spliit.onrender.com/](url)

- Add QR code generation with embedded logo for sharing groups
  - Display QR codes in share dialog with Spliit logo overlay
  - Support downloading QR codes as PNG images
  - Use high error correction level for reliable scanning

- Add camera-based QR code scanner for joining groups
  - Enable mobile devices to scan QR codes directly
  - Toggle between URL input and QR scanning on mobile
  - Handle camera permissions and error states gracefully

- Add new dependencies:
  - qrcode.react (^4.2.0) for QR code generation
  - html5-qrcode (^2.3.8) for camera-based scanning

- Add translation keys for QR code UI elements
  - Share QR code dialog (title, description, download button)
  - Scanner mode toggle (URL mode vs QR mode)
  
 Generating a QR code can be done on desktop and mobile clients, using the same workflow as for sharing a URL, but with an additional QR code icon. Selecting this will generate the QR Code. 
 
 Generate QR Code icon
generate1

Show QR Code
generate2

For a new group member that is all that would be needed to take the user to the site and access the group. For existing Spliit users, there is also an option to aquire the URL via the QR Code and add it to their list. To go this go to the top level groups section , and select the current Add by URL button. If on a mobile device it will now offer the user to add by URL as per the current option, or via QR code, at which point it would prompt the user to 

Aquire Group via QR Code
aquire1


theo-engels:
removing in preference of PR #500

Issue #378, opened 2025-07-02
Add variable for OpenAI endpoint


IITYWYBMAB:
The OpenAI API has become a standard for many inference tasks, it seems the endpoint is hard-coded in this project.  Being that there are many users with inference capabilities it would be nice to be able to utilize those without modifying the codebase.  I'd like to propose that the OpenAI endpoint become an easily controllable variable, which still defaults to the public endpoint, yet allows users to easily adapt the workflow that you've established to their own inference provider.

Thank you for the consideration.

Issue #350, opened 2025-06-02
Privacy Concern with Public S3 Requirement for Receipt OCR


gravelfreeman:
I'd like to raise a privacy concern regarding the following documentation note:

> This feature relies on OpenAI GPT-4 with Vision and a public S3 storage endpoint.

This effectively makes all personal receipts publicly accessible on the internet.

The OCR extraction works great, but I believe it would make more sense to replace bucket storage with local storage. I’m running over a hundred containers, all of which use local volumes, so this would integrate more securely and efficiently.

For context, I’m using OCR extraction with Taxhacker and Mealie and both offer similar functionality without this privacy issue, so I’m confident it’s achievable here too.

Thanks for the great work on the app.

Steeven9:
Not only for the OCR part, also the basic receipts uploading relies on a public S3 bucket (see [docs of the next-s3-upload library](https://next-s3-upload.codingvalue.com/setup#s3-bucket)).

The upload works (because it uses the given access key) but then the display won't load since it relies on the bucket being public:

<img width="1026" height="323" alt="Image" src="https://github.com/user-attachments/assets/7fc1c74f-cb44-4ce8-bf60-269cbb87e7bd" />

This is a huge privacy concern - if I self-host my data, I don't want it to be public!

I suggest checking out other S3 libraries - such as the official [minio client](https://www.npmjs.com/package/minio) - which support forwarding the access key headers for viewing as well

Issue #84, opened 2024-02-05
Set upper limit to attachment size


manuerwin:
The app seems to allow any size file to be uploaded, strongly suggest something like a setting that optionally limits the max attachment size and displays an error/warning if exceeded to:
a) conserve storage and
b) prevent bad actors

See screenshot of the 35MB sample image I was able to upload just now (I've since deleted the expense, however I'm unsure if the image has been deleted from S3?)
![screencapture-spliit-app-groups-expenses-edit-2024-02-05-19_02_33](https://github.com/spliit-app/spliit/assets/982158/1c135c41-7121-47bd-bc8d-c75c7fa3875a)

scastiel:
Agree! I found a nice way to add a limit for the file size when using next-s3-upload (the third package I used), but it requires a change in the library ([see this open PR](https://github.com/ryanto/next-s3-upload/pull/179)).

manuerwin:
Seems like that will work, for S3 which is great. 
However if the size check happened in app/window, then it would be applicable to *any* file provider right?

scastiel:
You’re right, I just pushed a check for the file size on the client side, which is nice for user experience anyway. But ultimately, we need to perform a better check, as for now nothing prevents a user from tweaking the code in the browser to disable the check.

manuerwin:
After a bit more thought, suggest perhaps a max limit of 2MB?
That should allow for fairly large images and certainly PDF’s etc IMO.
The math on 5MB X many expenses => large over the long term IMO

shynst:
> After a bit more thought, suggest perhaps a max limit of 2MB? That should allow for fairly large images and certainly PDF’s etc IMO. The math on 5MB X many expenses => large over the long term IMO

A setting you should probably be able to tweak in a .env file, instead of having to hardcode.

manuerwin:
Great point about the .env file, agree!

alfonzjanfrithz:
I tried to use the scan receipt straight from phone. my phone take it at 8mb. So it wouldnt work unless i trick it to snap it in whatsapp(that compress it first)

So, one idea is to do the compress before upload. This will massively reduce upload time, storage cost and potentially speed up any downstream image analysis.

Probably could do some resize in client side, convert to something like WebP. I know some image have metadata like location etc, idk how much, but it could contribute to file size i reckon. so if we can remove this kind of thing, that could probably reduce.

In an event client cant compress. Then only put the limit and do some validation above.

PR #499, opened 2026-01-29
Feature: Non Public Object Storage

Neirpyc:
###  Context & Use Case
I host this Spliit instance for upcoming group holidays. To keep our data private, the entire instance sits behind an **authenticated proxy**, ensuring only our friends can access the tool. 

**The Problem:** Currently, S3-uploaded documents (receipts/images) require the bucket to be public for them to render in the browser. This effectively bypasses the proxy's security, potentially exposing personal documents to the open web if a URL is leaked.

**The Fix:** This PR moves all S3 interactions to the server. The bucket can now be 100% private. Spliit fetches and streams the files server-side, meaning only users already authenticated through the proxy can see them.

###  Key Improvements
* **Privacy First:** Documents are no longer public. They are served via application routes (`/api/documents/[id]`).
* **Zero-CORS Setup:** Since the browser no longer talks directly to S3, we can remove all CORS configurations from the bucket.
* **Secure AI Analysis:** Enables server-side generation of temporary authenticated URLs for AI document extraction (tested with Scaleway) without ever making the file public.

#### Backend (The "Proxy" Logic)
* **`POST /api/documents`**: Receives multipart uploads, stores them in S3, and saves metadata (dimensions, filename) to the DB.
* **`GET /api/documents/:id`**: The "secure gateway." It verifies the document exists and streams the data directly from S3 to the authenticated user.

#### Client & UI
* **Frontend Refactor:** Replaced `next-s3-upload` dependencies with a lightweight, native implementation.
* **`getImageData`**: A new client helper that calculates image dimensions using browser APIs to prevent layout shift without extra libraries.
* **Receipt Flow:** The "Create from Receipt" button now triggers a server-side upload + AI extraction, keeping the entire pipeline off the public internet.

#### Infrastructure
* **`src/lib/s3.ts`**: Centralized S3 client factory that only initializes if env vars are present.
* **Flexible AI Env Vars**: Added support for `OPENAI_BASE_URL` and model overrides, allowing for easy swapping between OpenAI and local/Scaleway providers.

### Security Note
By default, the SDK creates objects with private ACLs. This PR ensures that even if someone gets hold of the raw S3 link, the file remains inaccessible without the app's internal credentials.


Neirpyc:
Note: addresses https://github.com/spliit-app/spliit/issues/350, https://github.com/spliit-app/spliit/issues/84 and https://github.com/spliit-app/spliit/issues/378

brknkfr:
Fantastic!

I just deployed spliit on our server and I was missing the possibility to use with our [LocalAI](https://github.com/mudler/LocalAI) instance and I thought, that it would be nice to use private S3 buckets as we use [Garage](https://garagehq.deuxfleurs.fr/) as our S3 bucket provider which is a little bit complicated to configure to provide files over the web (different domains for the S3 endpoint and the web endpoint).

Now there is your PR which solves those two problems. The PR works like a charm (together with LocalAI and Garage).

Please include this in the next release, @scastiel.

src/app/groups/[groupId]/expenses/create-from-receipt-button-actions.ts
```diff
@@ -1,18 +1,52 @@
...
+}
+
+export async function extractExpenseInformationFromImage(id: string) {
   'use server'
+  if (!env.NEXT_PUBLIC_ENABLE_RECEIPT_EXTRACT)
+    throw new Error('Receipt extraction is not enabled')
+
   const categories = await getCategories()
 
+  const resolvedUrl = await resolveDocumentToPresignedUrl(id)
```
Comment (qvalentin) [inline src/app/groups/[groupId]/expenses/create-from-receipt-button-actions.ts:45]: OpenAI also supports sending images via base64 encoded urls: https://developers.openai.com/api/docs/guides/images-vision?format=base64-encoded

This would make this way easier as we never need a third-party (OpenAI) to download the image.

Comment (Neirpyc) [reply src/app/groups/[groupId]/expenses/create-from-receipt-button-actions.ts:45]: It’s definitely possible. That said, it would make requests heavier due to the base64 encoding overhead, and it would shift more load onto the server since it would need to read and encode the file instead of offloading that to S3. Also, presigned URLs are already widely supported, so there isn’t really a compatibility concern.

Given that, I think presigned URLs are still a better fit for this use case for now. That said, if there’s a concrete need for base64, I could revisit this or make it a configurable option.

brknkfr:
This pull request doesn't apply anymore now. Will there be an update for this very useful pull request?

PR #526, opened 2026-06-11
Support tricount imports

Mubelotix:
Hi!

I have been hosting my own instance of Spliit for a long time and really appreciate the project. As I see that there is currently a high level of activity that can be a lot to handle for you, I wanted to offer my help. I would be more than happy to commit as a new maintainer to help with handling issues, reviewing MRs, and driving new updates to keep the project moving forward!

To start off, this MR implements a much-requested feature: importing groups and expenses. I implemented it for Tricount but it could be reused for splitwise.

#### What this MR does:

Adds a clean, interactive import section to the Group Creation page with step-by-step instructions guiding the user on how to obtain their data export under GDPR terms from Tricount.

It allows the user to select the group's main currency during the import. All imported expenses are converted to this currency using the rates in the CSV file (falling back to Frankfurter API if a rate is missing).

It also fixes an existing bug where the database field  originalAmount  was stored in minor units (cents) but was not converted to/from decimals in the frontend form ( submit  and  defaultValues ), causing 100x scaling display issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ Spliit is a free and open source alternative to Splitwise. You can either use th
 - [x] Search for expenses in a group [(#51)](https://github.com/spliit-app/spliit/issues/51)
 - [x] Upload and attach images to expenses [(#63)](https://github.com/spliit-app/spliit/issues/63)
 - [x] Create expense by scanning a receipt [(#23)](https://github.com/spliit-app/spliit/issues/23)
+- [x] Import group and expenses from Tricount CSV export
 
 ### Possible incoming features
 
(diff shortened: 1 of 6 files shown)


antonio-ivanovski:
Thanks for working on this! I already support importing from the original Spliit and Splitwise in my fork, and I'd love to see Tricount import there as well.

If you're interested in continuing the work, I'd be happy to review a PR and help get it merged.

t0ma5:
In `importTricount.procedure.ts`, the per-expense lookup that scans every CSV row for `targetCurrencyCode` is O(n²). Building a `Map<currency, rate>` once before the expense loop (first non-NaN rate per code) keeps the same fallback to Frankfurter and avoids that on large GDPR exports.

Happy to send a follow-up patch if useful — we used that pattern in our Workers/D1 fork.

PR #530, opened 2026-06-23
Add fixed monthly expense grouping setting

Dustin-ONeil:
## Summary

Adds an opt-in group-wide setting to display expense date headers as fixed calendar-month buckets instead of the existing relative buckets.

By default, groups continue using the current behavior with labels like “This week”, “Earlier this month”, and “Last month”. When enabled, expenses are grouped by calendar month, which is more useful for recurring monthly expenses, roommates, and shared bills.

## Changes

- Added `fixedExpenseDateGroups` to `Group` with a default of `false`
- Added a Prisma migration for the new field
- Added a separate “Group settings” section in the group form
- Kept the existing “Local settings” section and Active user selector unchanged
- Extracted expense date grouping into a small helper module
- Added fixed calendar-month grouping using UTC date parts from `expense.expenseDate`
- Kept relative grouping behavior unchanged when the setting is disabled
- Added tests for fixed monthly labels and UTC month handling

## Testing

- `npm test`
- `npm run check-types`
- `npm run lint`
- Targeted Prettier check on touched files

## Transparency Note

This is my first request using Codex 5.5 at highest enterprise setting for implementation help. I’m trying it out and I want to be transparent so reviewers can flag anything that feels off or low-quality.

### DIFF
--- a/messages/en-US.json
+++ b/messages/en-US.json
@@ -38,7 +38,8 @@
       "lastMonth": "Last month",
       "earlierThisYear": "Earlier this year",
       "lastYear": "Last year",
-      "older": "Older"
+      "older": "Older",
+      "currentMonth": "Current Month"
     }
   },
   "ExpenseCard": {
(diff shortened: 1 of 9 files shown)


Dustin-ONeil (commented):
Reviewed for common GPT-related issues and unnecessary bloat. Let me know whether this feels lightweight enough and covers the right use cases. I'm happy to redirect the implementation as needed

PR #531, opened 2026-06-24
Fix bug when pasting currency amount (parsing commas and decimals)

Dustin-ONeil:
## Summary

Fixes pasted currency values being misread when they include thousands separators.

For example, `-$1,659.84` was previously normalized as `-1.65984` because the first separator was treated as the decimal separator. This change normalizes common US and European pasted number formats into the existing internal format: digits, optional leading `-`, and `.` as the decimal separator.

## Changes

- Added a small `normalizeNumberInput` helper
- Replaced the old expense form currency sanitizer
- Supports common pasted values like:
  - `-$1,659.84`
  - `-€1.659,84`
  - `1,234,567.89`
  - `1.234.567,89`
- Keeps simple decimal entry working for both `1.23` and `1,23`
- Applies the same normalization to amount, original amount, conversion rate, and split-by-amount fields
- Does not add settings, migrations, or UI changes

## Testing

- `npm test`
- `npm run check-types`
- `npm run lint`
- Targeted Prettier check on touched files

## Transparency Note

This is my second small request using Codex for implementation help. I’m trying it out to see whether it improves my workflow, and I want to be transparent so reviewers can flag anything that feels off or low-quality.

### DIFF
--- a/src/app/groups/[groupId]/expenses/expense-form.tsx
+++ b/src/app/groups/[groupId]/expenses/expense-form.tsx
@@ -38,6 +38,7 @@ import { randomId } from '@/lib/api'
 import { defaultCurrencyList, getCurrency } from '@/lib/currency'
 import { RuntimeFeatureFlags } from '@/lib/featureFlags'
 import { useActiveUser, useCurrencyRate } from '@/lib/hooks'
+import { normalizeNumberInput } from '@/lib/number-input'
 import {
   ExpenseFormValues,
   SplittingOptions,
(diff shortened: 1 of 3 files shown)


Dustin-ONeil (commented):
Reviewed for common GPT-related issues and unnecessary bloat. Let me know whether this feels lightweight enough and covers the right use cases. I’m happy to add more robust edge cases or stronger string-handling scenarios as needed.

PR #555, opened 2026-08-13
Add monthly category spending visuals (v1)

t0ma5:
## Summary
- Clean v1 of monthly spending analytics: **stacked monthly chart + category breakdown only** (balance timeline deferred per @quintushr's review on #532).
- Builds on @Dustin-ONeil's work in https://github.com/spliit-app/spliit/pull/532 — thank you!
- Addresses https://github.com/spliit-app/spliit/pull/532#issuecomment-5275743269:
  - Dropped balance timeline (and unused timeline payload/fields) for v1
  - Split the large UI into reviewable modules (palette, stacked chart, legend, breakdown, controls, utils)
  - `stats.get` monthly params are optional — Totals no longer triggers monthly aggregation / a second heavy payload
  - Category labels go through next-intl `Categories` messages; timeline-only English strings removed
  - Prettier applied to the changed files

cc @quintushr @Dustin-ONeil

## Test plan
- [ ] Open Stats on a group with expenses across multiple months/categories
- [ ] Confirm stacked bars/columns + category breakdown render and update with range/grouping controls
- [ ] Toggle dark mode — chart segments/labels remain readable (no hard-coded white chart chrome from timeline)
- [ ] Switch locale — MonthlySpending strings + category names are translated (en-US keys; others fall back via deepmerge)
- [ ] Confirm Totals card still loads without requiring monthly params
- [ ] Create/edit/delete an expense and confirm stats invalidate/refresh
- [ ] `npm test -- --testPathPattern=monthly-spending`
- [ ] `npm run check-formatting`

Made with [Cursor](https://cursor.com)


BastiOfBerlin:
Hi,
thanks for you PR. Could you rebase the PR on main after #584 landed?

t0ma5:
Rebased onto current `main` after #584 (`feat: expand group stats`).

**Conflicts** were in `messages/en-US.json`, `stats/page.client.tsx`, and `stats/get.procedure.ts` (deleted on main). Resolution kept #584's architecture: period selector, `groups.stats.overview` as the single page query, and the existing Summary / Spending over time / by participant / by category / Recurring / Totals cards.

**Kept (unique vs #584)**
- Stacked monthly category chart (bars/columns), palette, legend, grouping + 3/6/12/all window, round-amounts, and category breakdown
- Chart-local grouping/range applied client-side from the overview payload so those controls do not refetch the other cards
- `utils.groups.stats.invalidate()` on expense create/edit/delete

**Dropped as duplicate / obsolete**
- Restoring `groups.stats.get` (overview already loads expenses once)
- A second totals-only stats fetch
- Replacing #584's spending-over-time or by-category cards

New head: `bc7f3d1` on `stats-monthly-category-v1`.

quintushr:
Thanks @t0ma5, for this — the code itself is clean: well-typed, unit-tested, logic properly split
between `src/lib/monthly-spending.ts` and the rendering components, and the `groups.stats`
invalidation on expense create/edit/delete is a nice fix along the way. `tsc`, `eslint`,
`prettier` and `jest` all pass on the branch.

A few things I'd like to see addressed before merging:

**1. The card's range control silently composes with the page-level one.**
`getMonthlyCategorySpending` runs on `expenses` in `overview.procedure.ts:63`, i.e. the list
*already* filtered by `filterExpensesByDateRange`. So the card's "All" doesn't mean "all
history", and picking "Last 3 months" at the page level + "12 months" in the card shows 3
months with no explanation. Two visible period selectors that multiply into each other is
confusing — either compute this card from `allExpenses`, or drop the local selector and let
it follow the page range like the other cards do.

**2. The window is anchored on the last expense, not on today.**
`applyMonthlySpendingView` takes the last month *that has data* (`monthly-spending.ts:152`).
A group inactive for a year shows "6 months" of last year's months, labelled `Jan`, `Feb`…
with no year (`stacked-chart.tsx:112` and `:330` format without `year`). In "All" mode
spanning several years the same labels repeat. Worth anchoring on the current month, and
including the year in the labels whenever the window crosses one.

**3. Overlap with the existing cards.**
The page already has "Spending over time" (monthly bars) and "By category" (category bars
with %). `MonthlyCategoryBreakdown` is a third rendering of the latter, with a possibly
different time window, and unlike the other two it has no `ExpensesDialog` drill-down. Could
this card either replace "Spending over time", or drop its breakdown section in favour of the
existing "By category"?

**4. The palette: mostly dead code, and it diverges from the existing convention.**
`getCategoryColorFor` only ever returns `palette[0]` (`category-palette.ts:283`), so every
color at index ≥ 1 in `CATEGORY_COLORS_BY_GROUP` and `FALLBACK_CATEGORY_COLORS` is
unreachable — roughly 150 of the file's 284 lines. The rest of the stats page colors bars
with `hsl(var(--chart-N))` (`stat-bar.tsx`, `category-breakdown.tsx`); this adds a second,
hardcoded Tailwind system, plus six hand-picked overrides that only cover `Utilities`. And in
"Detailed" mode, 44 categories map onto 12 colors via `categoryId % 12`, so adjacent segments
will collide. Reusing the `--chart-*` variables would shrink this a lot.

**5. "Columns" + "All" isn't readable.**
`stacked-chart.tsx:210` lays out `repeat(N, minmax(0, 1fr))` with no horizontal scroll; at
40–60 months the columns are a couple of pixels wide. The bars view likewise renders one row
per month with no cap. Some scrolling or a cap on the number of rendered months would help.

Thanks !

t0ma5:
Thanks @quintushr — addressed in f18b1cd:

1. **Range composition.** Dropped the card-local 3/6/12/All selector. The stacked chart now follows the page period, like the other cards. Grouping (Categories / Detailed) stays local so it does not refetch.

2. **Anchor on today.** The month window ends at the current month when the page range includes today (inactive groups show trailing empty months). A past custom range still stops at its `to` date. Axis labels include the year whenever the window crosses one.

3. **Overlap.** Removed the card’s category breakdown. Spending over time (with drill-down) and By category stay as they are; this card is only the stacked monthly view.

4. **Palette.** Replaced the unused Tailwind palettes with `hsl(var(--chart-N))`, assigned by display order so adjacent segments do not share a color until the 6th category.

5. **Columns + All.** Columns use a fixed min width and horizontal scroll; bars cap height and scroll vertically.

`npm test -- --testPathPattern=monthly-spending` passes (9 tests).

scastiel (approved):
Thanks @t0ma5 — all five points from @quintushr's review are addressed, and the branch passes tsc / eslint / prettier / jest for me as well. I also ran it locally (light/dark, bars/columns, Categories/Detailed, phone width) and it looks good.

Two small things I noticed in the Columns view, fine as a follow-up:
- When there are more months than fit, the scroller starts at the left, so the most recent months are hidden until you scroll. Defaulting the scroll position to the end would help.
- The amount label above the tallest column is clipped by the `overflow-x-auto` container (it renders above the `h-48` box), so the max month never shows its total.

Approving — thanks for the thorough follow-up!

t0ma5:
Follow-up for the two Columns notes is in **#645**: scroller starts at the latest months, and the tallest column's amount stays visible (label is in layout instead of sitting above the `h-48` box).

Issue #553, opened 2026-08-13
Tracking: coordination of merging forks


BastiOfBerlin:
I've been running a fork of Spliit for a while and, now that I have maintainer
access here, I'd like to contribute the work back. It's a lot — ~90 commits
across ~125 files — so rather than one unreviewable PR, this issue tracks a
sequenced series of small ones.

Nothing below lands without normal review. If any individual item isn't wanted
upstream, it just stays in the fork; the series is designed so each PR can be
dropped without blocking the others.

---

## ✅ The urgent one is closed

**`Expense.originalAmount` was holding two different units on `main`** — #425
converted the expense form to write minor units without a data migration, so
rows written before each instance deployed it were still in major units and
displayed 100× too small. #550 merged the backfill, so this is done.

<details>
<summary>My fork had a version of this migration too. #550's is better and mine isn't going anywhere near upstream.</summary>

Mine does a blanket `originalAmount * 100`, which assumes every row predates the
cutover — so it would **double-convert every row written after #425 shipped**.
It also hardcodes ×100 rather than 10^decimals, making it wrong for BHD, KWD,
JOD, OMR and TND regardless. #550 classifies each row by its own values instead,
which is the only thing that works when the cutover happens at each
self-hoster's deploy time.
</details>

---

## ✅ Phase 1 — merging existing community PRs — done

Several open PRs here were already carried in my fork and had been running in
production for months. These were merged as the originals so the authors keep
their attribution, before layering anything of mine on top. All eight are in.

- [x] #515 — `frankfurter.app` → `frankfurter.dev/v1` (currency conversion was
      broken on `main`)
- [x] #516 — currency data
- [x] #505 — Docker Compose documentation
- [x] #491 — restore keyboard navigation in the category and currency selectors
- [x] #500 — QR code in the share popover, and a camera scanner for joining
      (landed as #558)
- [x] #486, #507, #521 — add COP, MKD, MYR
- [x] #503 — enable Next 16's `reactCompiler`
- [x] #219 — `build-image.sh`: POSIX parsing instead of `node -p`, so the image
      builds on hosts without Node

**#500 needed a rebase**, which went in as #558. @theo-engels' commit was
cherry-picked byte-identical; what conflicted was only the lockfile, whose diff
was cut from `d3b151e` back in January, so I regenerated it against `main`
instead of replaying it — after #556 that seemed the safer habit.

**Two things from #558 are still open**, both flagged in its body:

[...]

PR #559, opened 2026-08-13
fix: honour locale week start when grouping expenses and activity

BastiOfBerlin:
# fix: honour locale week start when grouping expenses and activity

First of the series in #553. Self-contained, ~100 lines, fully unit-tested — no
dependency on anything else queued there.

## The bug

The expenses and activity lists group entries under time-period headers ("This
week", "Earlier this week", "Last week", …). Week boundaries are computed with
dayjs' `isSame(date, 'week')`, which reads dayjs' **global** locale `weekStart`.
That global is never wired to the app's active next-intl locale, so it stays at
its default of Sunday for every user, in every locale.

The visible effect: a Sunday entry is grouped into the current week even for the
many supported locales whose week starts on Monday. On a Monday, a `de-DE` user
sees yesterday's expense filed under "This week" rather than "Last week" —
Sunday is last week for them.

Spliit ships far more Monday-first locales than Sunday-first ones, so the
default is wrong for most of the user base.

## The fix

A new `src/lib/date-groups.ts`:

- `getWeekStartsOn(locale)` reads the first day of the week from `Intl.Locale`
  week info (`getWeekInfo()`, falling back to the `weekInfo` property), and
  converts it from the ECMA-402 1..7 Monday-first convention to dayjs' 0..6
  Sunday-first one. Locales without week info, or an unparseable locale, fall
  back to Monday.
- `isSameWeek(a, b, weekStartsOn)` compares the two dates' week starts.

The helper takes the first day as an argument rather than reading dayjs' global,
which keeps it deterministic and testable. The two list components pass what
they get from next-intl's `useLocale()`.

## Notes for review

- **Nothing configures dayjs globally.** I deliberately did not call
  `dayjs.updateLocale()` — a global mutation would fix these two call sites and
  silently change every other date computation in the app.
- **Why Monday as the fallback**, not Sunday: it is correct for the majority of
  the locales Spliit ships. It only applies on engines without `Intl` week info,
  which in practice means very old runtimes.
- `activity-list.tsx` gains one extra call: the "Last week" bucket also had to
  move to `isSameWeek`, since `today.subtract(1, 'week').isSame(date, 'week')`
  has the same global-locale problem.

## Verification

`npm ci --ignore-scripts`, `npx prisma generate`, `check-types`, `lint`,
`check-formatting` all clean against current `main` (lint: 16 warnings, all
pre-existing). `date-groups.test.ts` covers `de-DE`, `en-US`, the unparseable
fallback, both week conventions across a Monday/Sunday boundary, and
time-of-day insensitivity.

`npm test` is not wired into CI yet — a later PR in the series adds that.

### DIFF
--- a/src/app/groups/[groupId]/activity/activity-list.tsx
+++ b/src/app/groups/[groupId]/activity/activity-list.tsx
@@ -4,9 +4,10 @@ import {
   ActivityItem,
 } from '@/app/groups/[groupId]/activity/activity-item'
 import { Skeleton } from '@/components/ui/skeleton'
+import { getWeekStartsOn, isSameWeek } from '@/lib/date-groups'
 import { trpc } from '@/trpc/client'
 import dayjs, { type Dayjs } from 'dayjs'
-import { useTranslations } from 'next-intl'
+import { useLocale, useTranslations } from 'next-intl'
 import { forwardRef, useEffect } from 'react'
 import { useInView } from 'react-intersection-observer'
 import { useCurrentGroup } from '../current-group-context'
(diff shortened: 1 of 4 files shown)


watou:
I'm running [this 1.22.0 docker image](https://hub.docker.com/layers/ddivervs/spliit2/1.22.0/images/sha256-adaa2a7d9791d4cd17b8196b1badccb37ebe14460a1917fce51f598f844b3848) in and accessing it in Ireland, a country where the week starts on Monday, and I see entries under the "This Week" heading from Sunday, August 16.  This occurs on Google Chrome 151.0.7922.138 and  151.0.7922.170, and also Zen browser 1.21.14b (Firefox 153.0.4).  All my locale settings are consistent with my actual locale and specifies Monday as the first day of the week.

PR #582, opened 2026-08-14
feat: add new locale labels for Estonian, Lithuanian, Hungarian, and Swedish

daunera:
This commit enhances the localization support by adding new language labels for Estonian (et), Lithuanian (lt), Hungarian (hu), and Swedish (sv) to the localeLabels object in the i18n request module. This update improves accessibility for users in these language regions.


daunera:
@scastiel This pr only add the options to use already weblate translated languages in the app. Recommend to merge

PR #584, opened 2026-08-16
feat: expand group stats

BastiOfBerlin:
# feat: expand group stats

Part of the series in #553. The largest feature PR in it — 19 files, ~1470
lines, three commits.

## What lands

The stats tab currently shows one Totals card. This adds five more, and a period
selector (all time / this month / last 30 days / this year) that scopes all of
them:

| Card | Shows |
| --- | --- |
| Summary | Expense count, average, largest, active span |
| Spending over time | Total per month |
| Spending by participant | What each paid, how many expenses, and their share |
| Spending by category | Total per category |
| Recurring expenses | Estimated monthly and yearly cost of active series |

## Three commits, in review order

**1. The aggregation layer** (`totals.ts`, `api.ts`) — pure functions over the
expense list, with 13 new unit tests. **This commit stands alone: `check-types`
is clean and all 114 tests pass with no UI in the tree**, so the arithmetic can
be reviewed on its own before looking at a single component.

Reimbursements are excluded throughout. Unlike balances, spending stats
deliberately ignore settling up.

`getSpendingByParticipant` splits each expense **once** through `lib/shares` and
hands out the results, rather than re-deriving the whole split per participant.
That also means it agrees with the balances tab by construction — the thing #562
was about.

**2. The page** — the six cards, the period selector, chart colour variables for
both themes, and `en-US` strings only so Weblate picks them up normally.

It replaces `get.procedure` with a single `groups.stats.overview`. Six cards
each fetching their own slice meant six round trips over the same expense list;
the loader reads it once and every card takes a prop. That consolidation was a
follow-up commit in my fork — it's folded in here rather than shipped as
"add six queries" then "make it one".

**3. E2E coverage** — see below. Separate so you can drop it.

## One bug fixed on the way in

`getActiveRecurringExpenses` filters on the *open* `recurringExpenseLink`
(`nextExpenseCreatedAt: null`). Every materialised frame of a recurring series
keeps its own link, so counting frames counts one subscription many times — a
daily expense running for a month reported ~30×. Filtering on the open link
counts each series once. That was a separate fix in my fork; it's folded into
the commit it fixes.

## Verification

Against `c23273c`, Node 24 / npm 11: `npm ci --ignore-scripts`,
`npx prisma generate`, `check-types`, `lint` (16 warnings / 0 errors, all
pre-existing), `check-formatting`, `npm test` — **7 suites, 114 tests** (101
before) — and a full `npm run build`.

**And I ran your E2E suite: 32/32 passing.** That includes `stats.spec.ts`,
which is the one this PR could most plausibly have broken, since it rewrites
`totals.tsx` and the whole stats page. The `Stats.Totals` strings it asserts on
are unchanged.

The third commit adds `e2e/stats-breakdown.spec.ts` — the participant and
category breakdowns, the summary figures, and the period selector actually
restricting what the cards report. Mutation-checked rather than assumed, per
your note in #577:

```
Expected substring: "$61.00"
Received string:    "SummaryKey figures for the selected period.Expenses2
                     Average expense$60.00Largest expense$90.00Dinner
                     Active periodAug 16, 2026"

Expected substring: "$21.00"
Received string:    "TotalsSpending summary of the entire group.
                     Total group spendings$20.00"
```

So the average is genuinely computed, and switching the period genuinely drops
the out-of-range expense (that test records one dated 2020-01-15 and one today,
then switches to "This year" — a fixed past date rather than a relative one, so
it holds whenever the suite runs).

E2E ran against a real Postgres 16 and the built app rather than the compose
stack — no Docker daemon where I work — driven through `E2E_BASE_URL`. Same app,
same specs; what it does not exercise is the image build or `scripts/e2e.sh`.

## Deliberately not in this PR

The date-range **custom** option and the category/month **drill-downs** (click a
bar to see the expenses behind it) are the next PR in the series. They build on
these cards, and folding them in would roughly double an already large diff.
That's why `Stats.range` here has no `custom`/`from`/`to` keys.

## Worth flagging separately

`e2e.yml` triggers on `workflow_dispatch` and `push: tags: ['*']`, so the E2E
suite doesn't run on pull requests — a PR that breaks it goes green and surfaces
when you cut a tag. Adding `pull_request` would close that at the cost of an
image build plus a browser run per PR. Your call on the CI minutes; happy to
send it as a one-liner, but I didn't want to slip it into an unrelated PR.


quintushr:
Hi @BastiOfBerlin ,

Here are the two PRs that were trying to add these stats. I think we could take some ideas from what was done there and add them to what has just been merged #532 and #555

PR #585, opened 2026-08-16
Vouch for the iOS app so group links open it

scastiel:
The web half of Universal Links. [spliit-app/spliit-ios#18](https://github.com/spliit-app/spliit-ios/pull/18) shipped the entitlement and the link parser; both are inert until the site names the app, and that naming is a single file which lives here.

```
GET /.well-known/apple-app-site-association
```

**Every instance serves it, not only spliit.app.** The app follows links to whichever instance it is pointed at, and iOS only hands a link over if that host vouches for the app — so a self-hosted instance gets working group links with nothing to configure.

## Why a Route Handler and not `public/`

iOS ignores the association without explanation unless it arrives as `application/json` on a plain 200, and the extensionless name gives a static host nothing to guess the type from — `application/octet-stream` is the usual result. A handler states the type outright. `dynamic = 'force-static'` keeps it prerendered at build time (`○ /.well-known/apple-app-site-association` in the build output).

The claim is `/groups/*`, which covers the sub-pages too. That is deliberate rather than sloppy: the app's parser reads the segment after `groups`, so a link to a group's balances opens that group instead of the app shrugging at a URL it was handed.

## Two tests, because they fail differently

- **Unit** — the payload and the content type, the parts iOS rejects silently.
- **Playwright** — whether the server answers at that path *at all*. `.well-known` is a dot-directory under `src/app`, so that is Next's file-system router's decision, not ours; a build that quietly stopped emitting the route would leave the unit test green and Universal Links dead.

## Verification

`next build` lists the route as static; against `next start` it returns `200`, `content-type: application/json`, no redirect. Full jest suite 117 passed; `tsc`, eslint and prettier clean.

The E2E workflow is dispatch/tag-triggered rather than automatic on PRs, so it was [dispatched on this branch](https://github.com/spliit-app/spliit/actions/runs/31955324046) to exercise the new spec against the Docker image — the artifact that actually ships. `.dockerignore` has no dot-glob, so the directory is in the build context.

## Still needed, outside this repo

Associated Domains has to be enabled for the App ID in the Apple developer portal — a signed device build fails until it exists (simulator builds and CI are unaffected). Until this PR is deployed *and* that capability is added, an https group link opens Safari exactly as it does today. Nothing regresses either way.

https://claude.ai/code/session_01QAPVbmCRTBMm7u8LDMyQPR

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,6 +167,12 @@ The application has a health check endpoint that can be used to check if the app
 - `GET /api/health/readiness` or `GET /api/health` - Check if the application is ready to serve requests, including database connectivity.
 - `GET /api/health/liveness` - Check if the application is running, but not necessarily ready to serve requests.
 
+## iOS app links
+
+Every instance serves `GET /.well-known/apple-app-site-association`, the file iOS reads to decide whether the site is allowed to hand `/groups/<id>` links to the [Spliit iOS app](https://github.com/spliit-app/spliit-ios) instead of opening them in Safari.
+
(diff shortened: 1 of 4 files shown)


BastiOfBerlin:
@scastiel : You might want to check #587 for iOS compatibility

BastiOfBerlin (approved):

PR #610, opened 2026-08-30
Cap Zod inputs for groups, expenses, and pagination

t0ma5:
## Summary
- Caps `information` (2000), participants (100), expense title/notes, document URLs, and paid-for rows.
- Caps `groups.list` to 100 ids, expense/activity `limit` to 100, and `forUser` to 100 groups.

Prevents oversized tRPC payloads from becoming an easy memory/DoS footgun. Existing valid groups and forms are unaffected.

## Test plan
- [ ] Create/edit a group with a long information field under 2000 characters; save succeeds.
- [ ] Expense list pagination still loads the next page.
- [ ] Sending `limit: 10000` on expenses.list is rejected by Zod.


scastiel (changes requested):
Thanks, good direction. A few things before merging:

1. **Home page can hang for users with many groups.** `groups.list` (`groupIds.max(100)`) and `balances.forUser` (`groups.max(100)`) receive the whole `recentGroups` localStorage list, which is never pruned. Both callers (`recent-group-list.tsx`, `global-balance-card.tsx`) render `isLoading || !data` with no error branch, so >100 stored groups → Zod rejection → spinner forever. Please slice/chunk on the client (or handle the error state).

2. **Missing i18n keys.** `information.max(2000)` and `notes.max(5000)` have no message key, so `FormMessage` shows Zod's raw default ("Too big: expected string to have <=2000 characters"). Please add `max2000` / `max5000` to `SchemaErrors` like you did for `max200`.

3. **Limits vs. existing data.** These columns are unbounded `text`, so any existing row above a cap becomes uneditable until truncated. 200 (title) and 5000 (notes) look fine, but 2000 for group `information` feels tight (rules, bank details, itineraries…) — I'd go with 10000.

Minor:
- ID cap is `max(30)` everywhere except `groups.list` (`max(64)`) — pick one.
- `documents.width/height.max(10000)`: a phone panorama (~16000px) fits under the 5 MB limit and would now be rejected; the dims only feed `<Image>` layout, so I'd drop that cap.
- `filter.max(200)`: the search input has no `maxLength`, so a long search errors instead of returning nothing — `maxLength={200}` on the input would cover it.

t0ma5:
Thanks for the review — all of this is addressed on the rebased branch.

GitHub would not reopen this PR (`t0ma5:pr/zod-input-caps` has no history in common with `spliit-app:main` after the rewrite), so the same work is in **#644**:

1. Home page slices `recentGroups` to 100 and shows an error/retry instead of hanging on `isLoading || !data`.
2. i18n keys for the information/notes caps (`max10000` / `max5000`).
3. Group `information` cap raised to 10000.

Minors too: IDs unified at 64, image `width`/`height` caps dropped, expense search `maxLength={200}`.

PR #611, opened 2026-08-30
Add error boundaries that do not leak exception messages

t0ma5:
## Summary
- Adds `src/app/error.tsx` and `src/app/groups/[groupId]/error.tsx`.
- Copy is generic i18n (`Errors.*`); `error.message` is never rendered.

Stops unhandled exceptions from leaking internals to anyone who can open a group URL.

## Test plan
- [x] Temporarily throw in a group page; the boundary shows the generic message and Retry works.
- [x] Confirm the thrown message string does not appear in the HTML.

### DIFF
--- a/messages/en-US.json
+++ b/messages/en-US.json
@@ -11,6 +11,12 @@
   "Header": {
     "groups": "Groups"
   },
+  "Errors": {
+    "title": "Something went wrong",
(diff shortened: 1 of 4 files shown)


quintushr:
Thanks @t0ma5 for yout pr !!

PR #612, opened 2026-08-30
Reject non-integer minor-unit amounts on expense writes

t0ma5:
## Summary
- Asserts `amount`, `originalAmount`, and `paidFor.shares` are integers before create/update.
- Complements the existing Hamilton/remainder share math so a float cannot sneak into a minor-unit column.

## Test plan
- [ ] Create a normal even-split expense; it still saves.
- [ ] Calling createExpense with `amount: 10.5` throws before Prisma writes.


BastiOfBerlin:
Like your efforts to harden security!
Please add some tests to prevent future regression for this new functionality. Thx!

PR #616, opened 2026-08-30
Load expense participant IDs without fetching every expense

t0ma5:
## Summary
- `getGroupExpensesParticipants` is used only to know which participants cannot be deleted on the group edit form.
- It previously called `getGroupExpenses()` with no page, which hydrates every expense (payers, paid-for, documents count) and also runs `createRecurringExpenses()` as a side effect.
- Replace that with two small `distinct` Prisma queries for `paidById` and `ExpensePaidFor.participantId`.

## Test plan
- [x] Open group settings on a group with expenses: participants who paid or were paid for still cannot be removed.
- [x] A participant with no expenses can still be removed.
- [x] Viewing group settings no longer depends on loading the full expense list (large groups should stay snappy).

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/lib/api.ts
+++ b/src/lib/api.ts
@@ -127,14 +127,24 @@ export async function deleteExpense(
 }
 
 export async function getGroupExpensesParticipants(groupId: string) {
-  const expenses = await getGroupExpenses(groupId)
+  const [payers, paidFor] = await Promise.all([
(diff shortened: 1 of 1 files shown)


quintushr:
Thanks @t0ma5 for your pr !

PR #638, opened 2026-09-12
Keep by-amount shares when reopening an expense

### LINKED ISSUE #621: Uneven Split by Amount Gets Reset upon Saving
## Repro Steps
1. Create an expense. Add 3 participants.
2. Chang split mode to "Unevenly -- by amount"
3. Unselect 1 participant. Set the other participants a random split amount that differ in value. 
4. Save expense

Upon viewing expense, it will revert back to even split amounts despite still staying in correct split mode. Split amounts are lost.

Can confirm this reproes at least on commit ae35cd6. Based off some testing I believe this issue has been present since 2024.

scastiel:
Fixes #621

## What was happening

With split mode *Unevenly – By amount*, an expense saved with one participant unchecked came back with an **even** split when reopened. The database was fine (the right amounts were stored); it was the edit form that rebalanced them on load, and saving from there persisted the even split.

The same thing happened to every converted (multi-currency) by-amount expense, even with all participants checked.

## Root cause

The per-row share inputs under "Paid for" are driven by the parent `paidFor` field, but were wrapped in `FormField`, whose `Controller` also registers the row path as a field of its own. react-hook-form's `register` writes that path into the form values:

- `paidFor[-1].shares` for an unchecked participant (`findIndex` → -1) plants a stray `-1` key on the `paidFor` array;
- `paidFor[N].originalAmount: undefined` for each row of a converted expense adds a key the row object didn't have.

Either makes `paidFor` compare unequal to its default value, so `getFieldState('paidFor').isDirty` was `true` on mount. The BY_AMOUNT auto-balance effect is gated on that flag precisely so it leaves stored amounts alone (#210) — with the flag spuriously set, it redistributed the total evenly as soon as the edit form opened.

## Fix

- `src/components/ui/form.tsx`: add `FormFieldScope`, which provides the `FormField` context (so `FormControl` / `FormMessage` keep working) without registering anything.
- `expense-form.tsx`: use it for the two row inputs, and compute the row's index once instead of six `findIndex`/`some` lookups. Behaviour of the inputs is unchanged.

## Tests

Two Playwright regressions (both fail on `main` with `Expected "60", Received "50.00"`):

- `split-modes.spec.ts`: the reported scenario — one participant unchecked, reopen, values kept; saving untouched keeps the balances.
- `multi-currency.spec.ts`: converted expense, all participants checked, reopen, values kept.

Also verified manually in the browser (both scenarios, plus the per-row error message and auto-balance after a real edit still work). `npm test`, `check-types`, `lint` (0 errors), `check-formatting` all pass; the full e2e suite passes against `next dev` except `pwa.spec.ts`, which by design only runs against a production build.

https://claude.ai/code/session_018qL4xRK41UBzKr6k8Rj2Hm


scastiel:
Follow-up commit 6ed7ad1 from the code review:

- the row now uses the hoisted `isSelected` for the checkbox and share label as well;
- `FormFieldScope` takes a plain `name: string` (the copied generics could never be inferred) and its comment names the real trigger — registration only pollutes the values, the next change event (e.g. a selector re-submitting its default on mount) is what recomputes `dirtyFields`;
- `useFormField`'s error mentions `FormFieldScope`;
- `form.test.tsx` now mirrors the scoped wiring and asserts the invariant directly: `paidFor` stays pristine across a pristine change event. With the old per-row `FormField` wiring the assertion fails with `{"paidFor":true}`, so the regression is caught by Jest, not only by the e2e suite.

Left for a separate change (pre-existing, not touched here): the auto-balance effect still runs on mount gated on `isDirty`; the repayment defaults produce `paidFor: [undefined]` without `?to=`; all row inputs share one DOM id.

PR #645, opened 2026-09-13
Scroll monthly columns to the latest month and keep the max amount visible

t0ma5:
## Summary
- Default the monthly Columns scroller to the latest months.
- Keep the tallest column's amount label in layout so `overflow-x-auto` no longer clips it.

Follow-up to #555 (already merged). Addresses https://github.com/spliit-app/spliit/pull/555#pullrequestreview-5188945456

## Test plan
- [ ] Stats, Columns view, more months than fit: the latest months are visible without scrolling left first.
- [ ] The month with the highest total still shows its amount above the column (light and dark).
- [ ] Bars view is unchanged.


scastiel (approved):
Thanks!

